PR #134, opened 2024-06-03
feat: verify receipts in tests/mainnet.rs

### LINKED ISSUE #133: Receipts
- [ ] API: Return `Vec<PevmExecutionResult>` with `PevmExecutionResult` being `ReceiptAndState` instead of `ResultAndState`. We can use Alloy's [`Receipt` type](https://github.com/alloy-rs/alloy/blob/main/crates/consensus/src/receipt/receipts.rs) for now. 
https://github.com/risechain/pevm/blob/fd107cdbb8949413102c007f6d23d86b0d48875e/src/pevm.rs#L46
NOTE: We'll eventually rename  `InMemoryAccount` to `Account` to replace REVM's `Account` type in `State`. The end goal is to not leak our underlying executor types (REVM's forked types currently) to library users.

- [ ] Testing: Replace `must_succeed` with an `Option<&Header>` which could be a real block header. When it is `Some`, calculate the receipts root, logs bloom, and gas used and assert against it. We then pass in `Some(block.header)` for the [mainnet blocks](https://github.com/risechain/pevm/blob/main/tests/mainnet.rs). 
https://github.com/risechain/pevm/blob/fd107cdbb8949413102c007f6d23d86b0d48875e/tests/common/runner.rs#L31-L34

Relevant code from Reth:
- https://github.com/paradigmxyz/reth/blob/99c951a8bdde3c1f733e569832a08c75b7183f61/crates/ethereum/evm/src/execute.rs#L106-L109
- https://github.com/paradigmxyz/reth/blob/99c951a8bdde3c1f733e569832a08c75b7183f61/crates/ethereum/consensus/src/validation.rs#L21-L33

(*) We can skip requests (upcoming hardfork) for now.

kien-rise:
Resolves #133


src/pevm.rs
```diff
@@ -42,8 +44,11 @@ pub enum PevmError {
     UnreachableError,
 }
 
-/// Execution result of PEVM.
-pub type PevmResult = Result<Vec<ResultAndState>, PevmError>;
+/// Execution result of a transaction
+pub type PevmExecutionResult = (Receipt, EvmState);
```
hai-rise, line 48: Looking back, `PevmTxExecutionResult` is a better name. Let's also turn this into a `struct`.

src/pevm.rs
```diff
@@ -196,16 +201,23 @@ pub fn execute_revm<S: Storage + Send + Sync>(
...
         .map(|(mutex, value)| {
-            let mut result_and_state = mutex.into_inner().unwrap().unwrap();
-            result_and_state.state.insert(
+            let ResultAndState { result, mut state } = mutex.into_inner().unwrap().unwrap();
+            state.insert(
                 beneficiary_address,
                 post_process_beneficiary(&mut beneficiary_account_info, value),
             );
-            result_and_state
+            cumulative_gas_used += result.gas_used() as u128;
```
hai-rise, line 214: Deduplicate code with `execute_sequential`. Even better,  if we have a function taking in an iterator of `ResultAndState` and returns an iterator or vector of `PevmTxExecutionResult`, then we can separate output conversion with beneficiary post-processing & sequential execution.

tests/ethereum/main.rs
```diff
@@ -143,28 +142,13 @@ fn run_test_unit(path: &Path, unit: &TestUnit) {
...
-                        ExecutionResult::Success {
-                            output: Output::Create(b, None),
-                            ..
-                        } => b == Bytes::new(),
-                        _ => false,
-                    });
+                    let (receipt, _state) = &exec_results[0];
+                    assert!(receipt.status);
                 }
                 // Special cases where REVM returns `Ok` instead of `Err` on unsupported features.
```
hai-rise, line 148: We're already skipping it in the error case, let's just bite the bullet for now:
```
// Skipping special cases where REVM returns `Ok` instead on unsupported features.
(Some("TR_TypeNotSupported"), Ok(_)) => {}
```

src/pevm.rs
kien-rise: fixed

tests/ethereum/main.rs
kien-rise: updated the comment

tests/ethereum/main.rs
```diff
@@ -143,28 +142,13 @@ fn run_test_unit(path: &Path, unit: &TestUnit) {
                 // EIP-2681
                 (Some("TR_NonceHasMaxValue"), Ok(exec_results)) => {
                     assert!(exec_results.len() == 1);
-                    assert!(match exec_results[0].result.clone() {
-                        ExecutionResult::Success {
-                            output: Output::Create(b, None),
-                            ..
-                        } => b == Bytes::new(),
```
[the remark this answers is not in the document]
kien-rise: How about this?

```
assert!(exec_results[0].state.values().all(|account| account.info.is_empty_code_hash()));
```

Cargo.toml
```diff
@@ -21,9 +21,12 @@ revm = { git = "https://github.com/risechain/revm", rev = "f473fb17f127a8ead2bcc
 # RPC Storage dependencies
 # TODO: Put these behind an RPC flag to not pollute the core
 # library with RPC network & transport dependencies, etc.
+alloy-consensus = { git = "https://github.com/alloy-rs/alloy", rev = "64feb9bc51c8021ea08535694c44de84222f474e" }
```
hai-rise, line 24: Let's move core types to where `alloy-primitives` is above. This section is for RPC Storage dependencies that we'll make optional in the future.

src/pevm.rs
```diff
@@ -42,8 +45,37 @@ pub enum PevmError {
...
-/// Execution result of PEVM.
-pub type PevmResult = Result<Vec<ResultAndState>, PevmError>;
+/// Execution result of a transaction
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct PevmTxExecutionResult {
+    /// Receipt of execution
+    pub receipt: Receipt,
+    /// State that got updated
+    pub state: EvmState,
```
hai-rise, line 55: Add a `// TODO: Use our own type to not leak REVM types to library users`.

src/pevm.rs
```diff
@@ -42,8 +45,37 @@ pub enum PevmError {
...
+}
+
+fn collect_receipts(inputs: Vec<ResultAndState>) -> Vec<PevmTxExecutionResult> {
```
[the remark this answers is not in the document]
hai-rise: A better function signature is:
```rust
fn transform_output(
    revm_results: impl IntoIterator<Item = ResultAndState>,
) -> Vec<PevmTxExecutionResult> {
```

- Names are clearer, especially when we'll transform REVM's state type into our state type.
-  `impl IntoIterator<Item = ResultAndState>` is a micro-optimization that doesn't require the caller to first collect REVM's results into an intermediate `Vec` in memory. `execute_revm` can then pipe the `execution_results.into_iter()...` thing directly into this function.

src/pevm.rs
```diff
@@ -42,8 +45,37 @@ pub enum PevmError {
...
 }
 
-/// Execution result of PEVM.
-pub type PevmResult = Result<Vec<ResultAndState>, PevmError>;
+/// Execution result of a transaction
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct PevmTxExecutionResult {
+    /// Receipt of execution
+    pub receipt: Receipt,
```
hai-rise, line 53: Add a:
```
// TODO: Consider promoting to `ReceiptEnvelope` if there is high demand

```

src/pevm.rs
```diff
@@ -42,8 +45,37 @@ pub enum PevmError {
...
+    let mut cumulative_gas_used: u128 = 0;
+    let mut outputs = Vec::new();
+
+    for ResultAndState { result, state } in inputs {
+        cumulative_gas_used += result.gas_used() as u128;
+        let receipt = Receipt {
+            status: result.is_success(),
+            cumulative_gas_used,
+            logs: match result {
+                ExecutionResult::Success { logs, .. } => logs,
```
hai-rise, line 68: Let's steal a clearer version from Reth: `result.into_logs()`.
https://github.com/paradigmxyz/reth/blob/99c951a8bdde3c1f733e569832a08c75b7183f61/crates/ethereum/evm/src/execute.rs#L199

src/pevm.rs
```diff
@@ -196,7 +228,8 @@ pub fn execute_revm<S: Storage + Send + Sync>(
     // to avoid "implicit" dependency among consecutive transactions that read & write there.
     // TODO: Refactor, improve speed & error handling.
     let beneficiary_values = mv_memory.consume_beneficiary();
-    Ok(execution_results
+
+    let results = execution_results
```
hai-rise, line 232: Let's avoid no-ops for now.

src/pevm.rs
```diff
@@ -42,8 +45,37 @@ pub enum PevmError {
...
+    pub receipt: Receipt,
+    /// State that got updated
+    pub state: EvmState,
+}
+
+fn collect_receipts(inputs: Vec<ResultAndState>) -> Vec<PevmTxExecutionResult> {
+    let mut cumulative_gas_used: u128 = 0;
+    let mut outputs = Vec::new();
+
+    for ResultAndState { result, state } in inputs {
```
hai-rise, line 62: Cleaner with `into_iter`.

tests/common/runner.rs
```diff
@@ -50,30 +53,100 @@ pub fn test_execute_revm<S: Storage + Clone + Send + Sync>(
...
     assert_execution_result(
-        pevm::execute_revm_sequential(storage.clone(), spec_id, block_env.clone(), txs.clone()),
-        pevm::execute_revm(storage, spec_id, block_env, txs, concurrency_level),
+        &pevm::execute_revm_sequential(storage.clone(), spec_id, block_env.clone(), txs.clone()),
+        &pevm::execute_revm(storage, spec_id, block_env, txs, concurrency_level),
         false, // TODO: Parameterize this
     );
 }
 
+fn calculate_receipt_root(receipt_envelopes: Vec<ReceiptEnvelope>) -> B256 {
```
hai-rise, line 62: A comment to some doc or reference implementation would be nice.

tests/common/runner.rs
```diff
@@ -50,30 +53,100 @@ pub fn test_execute_revm<S: Storage + Clone + Send + Sync>(
...
+    // create a BTreeMap (order must be ascending)
```
hai-rise, line 63: Minor but let's start comments in upper-case for consistency.

tests/common/runner.rs
```diff
@@ -50,30 +53,100 @@ pub fn test_execute_revm<S: Storage + Clone + Send + Sync>(
...
+
+    // calculate the MPT root
+    let mut hash_builder = alloy_trie::HashBuilder::default();
+    for (k, v) in entries {
+        hash_builder.add_leaf(alloy_trie::Nibbles::unpack(&k), &v);
+    }
+    hash_builder.root()
+}
+
+fn get_receipt_envelopes(
```
hai-rise, line 83: Minor but `build` is better than `get` here. Throughout Rust's standard library `get` functions usually return references to existing data -- truly the act of getting, not allocating and returning ownership to new data!

tests/common/runner.rs
```diff
@@ -50,30 +53,100 @@ pub fn test_execute_revm<S: Storage + Clone + Send + Sync>(
...
+        .enumerate()
+        .map(|(index, receipt)| {
+            let key_buffer = alloy_rlp::encode_fixed_size(&index);
+            let mut value_buffer = Vec::new();
+            receipt.encode_2718(&mut value_buffer);
+            (key_buffer, value_buffer)
+        })
+        .collect::<BTreeMap<_, _>>();
+
+    // calculate the MPT root
```
hai-rise, line 75: I think we can omit this comment.

tests/common/runner.rs
```diff
@@ -50,30 +53,100 @@ pub fn test_execute_revm<S: Storage + Clone + Send + Sync>(
...
+        hash_builder.add_leaf(alloy_trie::Nibbles::unpack(&k), &v);
+    }
+    hash_builder.root()
+}
+
+fn get_receipt_envelopes(
+    block: &Block,
+    tx_results: &[PevmTxExecutionResult],
+) -> Vec<ReceiptEnvelope> {
+    // get the receipts
```
hai-rise, line 87: I think we can omit this comment.

tests/common/runner.rs
```diff
@@ -50,30 +53,100 @@ pub fn test_execute_revm<S: Storage + Clone + Send + Sync>(
...
+    let receipt_iter = tx_results.iter().map(|tx| tx.receipt.clone());
+
+    // get the tx_type list
```
hai-rise, line 90: I think we can omit this comment.

tests/common/runner.rs
```diff
@@ -50,30 +53,100 @@ pub fn test_execute_revm<S: Storage + Clone + Send + Sync>(
...
+    let tx_type_iter = block
+        .transactions
+        .txns()
+        .unwrap()
+        .map(|tx| TxType::try_from(tx.transaction_type.unwrap_or_default()).unwrap());
+
+    // zip them to get the receipt envelopes
+    let receipt_envelope_iter =
```
hai-rise, line 98: We can remove this intermediate variable.

tests/common/runner.rs
```diff
@@ -50,30 +53,100 @@ pub fn test_execute_revm<S: Storage + Clone + Send + Sync>(
...
+        Iterator::zip(receipt_iter, tx_type_iter).map(|(receipt, tx_type)| match tx_type {
+            TxType::Legacy => ReceiptEnvelope::Legacy(receipt.with_bloom()),
```
hai-rise, line 100: Looking at this pattern we can:
- Put `tx_type_iter` first to align with the `match`.
- Move `with_bloom()`  to `receipt_iter` to make it less redundant here?

tests/common/runner.rs
```diff
@@ -50,30 +53,100 @@ pub fn test_execute_revm<S: Storage + Clone + Send + Sync>(
...
+    let receipt_iter = tx_results.iter().map(|tx| tx.receipt.clone());
+
+    // get the tx_type list
+    let tx_type_iter = block
+        .transactions
+        .txns()
+        .unwrap()
+        .map(|tx| TxType::try_from(tx.transaction_type.unwrap_or_default()).unwrap());
+
+    // zip them to get the receipt envelopes
```
hai-rise, line 97: Avoid inline comments that add little to the code. Prefer why comments over what comments, too.

tests/common/runner.rs
```diff
@@ -50,30 +53,100 @@ pub fn test_execute_revm<S: Storage + Clone + Send + Sync>(
...
+}
+
 // Execute an Alloy block sequentially & with PEVM and assert that
 // the execution results match.
 pub fn test_execute_alloy<S: Storage + Clone + Send + Sync>(
     storage: S,
     block: Block,
     parent_header: Option<Header>,
     must_succeed: bool,
+    must_check_receipts_root: bool,
```
hai-rise, line 115: `must_match_receipts_root` sounds better. Also, we can now remove `must_succeed` from here and `assert_execution_result`. This supersedes `must_succeed` as it needs to `unwrap` an `Ok` out fo the execution results to calculate the receipts root anyway.

tests/common/runner.rs
```diff
@@ -50,30 +53,100 @@ pub fn test_execute_revm<S: Storage + Clone + Send + Sync>(
...
-        must_succeed,
+    let sequential_result = pevm::execute(
+        storage.clone(),
+        block.clone(),
+        parent_header.clone(),
+        concurrency_level,
+        true,
     );
+    let parallel_result = pevm::execute(
+        storage.clone(),
```
hai-rise, line 126: We don't need to clone `storage` and `parent_header` here?

tests/common/runner.rs
```diff
@@ -50,30 +53,100 @@ pub fn test_execute_revm<S: Storage + Clone + Send + Sync>(
...
+        block.clone(),
+        parent_header.clone(),
+        concurrency_level,
+        false,
+    );
+    assert_execution_result(&sequential_result, &parallel_result, must_succeed);
+
+    if must_check_receipts_root {
```
hai-rise, line 134: I don't mind having extra information in code, but this is not that interesting to Parallel EVM, and can be shortened to something like:
```rust
    // We can only calculate the receipts root from Byzantium.
    // Before EIP-658 (https://eips.ethereum.org/EIPS/eip-658), the
    // receipt root is calculated with the post transaction state root,
    // which we doesn't have in these tests.
    if must_match_receipts_root && block.header.number.unwrap() >= 4370000 {
        assert_eq!(
            block.header.receipts_root,
            calculate_receipt_root(get_receipt_envelopes(&block, &sequential_result.unwrap()))
        );
    }
```

Also, looking at this, we can call `get_receipt_envelopes` from inside `calculate_receipt_root` instead of chaining it here.

tests/common/runner.rs
```diff
@@ -50,30 +53,100 @@ pub fn test_execute_revm<S: Storage + Clone + Send + Sync>(
...
+    // calculate the MPT root
+    let mut hash_builder = alloy_trie::HashBuilder::default();
+    for (k, v) in entries {
+        hash_builder.add_leaf(alloy_trie::Nibbles::unpack(&k), &v);
+    }
+    hash_builder.root()
+}
+
+fn get_receipt_envelopes(
+    block: &Block,
```
hai-rise, line 84: Minor but we don't need the whole `Block` for this and having `txs` and `tx_results` as names match pretty well.

tests/ethereum/main.rs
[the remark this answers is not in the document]
hai-rise: This case is not very intuitive so let's put a comment behind:
```
// This is overly strict as we only need the newly created account's code to be empty.
// Extracting such account is unjustified complexity so let's live with this for now.
```

tests/mainnet.rs
```diff
@@ -14,10 +14,10 @@ pub mod common;
 
 #[test]
 fn mainnet_blocks_from_rpc() {
-    let rpc_url: Url = std::env::var("RPC_URL")
-        .unwrap_or("https://eth.llamarpc.com".to_string())
-        .parse()
-        .unwrap();
+    let rpc_url = match std::env::var("RPC_URL") {
+        Ok(value) if !value.is_empty() => value.parse().unwrap(),
```
hai-rise, line 18: When and where it would exist but be empty? We need practical cases to justify more code & logic.

tests/mainnet.rs
```diff
@@ -43,7 +43,7 @@ fn mainnet_blocks_from_rpc() {
             .unwrap();
         let rpc_storage = RpcStorage::new(provider, BlockId::number(block_number - 1));
         let db = CacheDB::new(&rpc_storage);
-        common::test_execute_alloy(db.clone(), block.clone(), None, true);
+        common::test_execute_alloy(db.clone(), block.clone(), None, true, true);
```
hai-rise, line 46: After removing `must_succeed` we can keep only one `true` here.

tests/raw_transfers.rs
```diff
@@ -102,5 +102,6 @@ fn raw_transfers_independent_alloy() {
         },
         None,
         true,
```
hai-rise, line 104: After removing `must_succeed` we can keep only one `false` for everything below from here.

Cargo.toml
kien-rise: > are these just dev-dependencies at the moment?

yes, all the added deps are for `tests/`

src/pevm.rs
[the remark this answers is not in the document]
kien-rise: Using `impl IntoIterator` is a good point, it follows [C-GENERIC](https://rust-lang.github.io/api-guidelines/flexibility.html#functions-minimize-assumptions-about-parameters-by-using-generics-c-generic).

src/pevm.rs
```diff
@@ -42,8 +45,37 @@ pub enum PevmError {
...
+pub struct PevmTxExecutionResult {
+    /// Receipt of execution
+    pub receipt: Receipt,
+    /// State that got updated
+    pub state: EvmState,
+}
+
+fn collect_receipts(inputs: Vec<ResultAndState>) -> Vec<PevmTxExecutionResult> {
+    let mut cumulative_gas_used: u128 = 0;
+    let mut outputs = Vec::new();
```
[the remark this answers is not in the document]
kien-rise: > no method named `size_hint` found for type parameter `impl IntoIterator<Item = ResultAndState>` in the current scope

let's use `.into_iter()` and `.map`

tests/mainnet.rs
kien-rise: Without this change, CI always fail when I make any PRs.

https://github.com/risechain/pevm/actions/runs/9108887873/job/25040598978?pr=78

tests/common/runner.rs
kien-rise: `must_succeed` seems to be still needed, doesn't it?

see:
```
fn raw_transfers_independent_alloy()
fn empty_alloy_block()
fn one_tx_alloy_block()
```

kien-rise: or do you mean that we should always treat it as `must_succeed = true` and hence `must_succeed` is not needed?

tests/common/runner.rs
kien-rise: changed to:

```
fn build_receipt_envelopes(
    txs: &BlockTransactions<Transaction>,
) -> Vec<ReceiptEnvelope>
```

tests/mainnet.rs
kien-rise: related to https://github.com/risechain/pevm/pull/134#discussion_r1627930717

Cargo.toml
hai-rise: Then please move them to `[dev-dependencies]`.

src/pevm.rs
```diff
@@ -42,8 +45,37 @@ pub enum PevmError {
     UnreachableError,
 }
 
-/// Execution result of PEVM.
-pub type PevmResult = Result<Vec<ResultAndState>, PevmError>;
+/// Execution result of a transaction
+#[derive(Debug, Clone, PartialEq, Eq)]
```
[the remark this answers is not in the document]
hai-rise: Blindly adding more code is always bad. Let's keep the `derive`s as minimal as possible then revisit it in a dedicated PR where we:
- Address all the `// TODO: Better types & API please`, `// TODO: Properly type this`, etc.
- Precisely decide what is `pub` then trace from there what should have eager derives.

After having such pre-Alpha API cleanup we can decide on a list of traits that all `pub` should derive.

src/pevm.rs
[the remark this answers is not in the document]
hai-rise: By "automatically" I mean `into_iter()` will use `size_hint()` internally when it collects a `Vec` for efficient allocation, so we don't have to use `Vec::with_capacity` ourselves.

src/pevm.rs
hai-rise: I find the latter readable but I guess that's subjective.

If we were to use intermediate variables for readability, though, the most valuable addition is the name.
`let results = execution_results.into_iter()...` or
`let execution_results = execution_results.into_iter()...` sound like missed opportunities.

We can settle for `let fully_evaluated_results = ` or `let fully_resolved_results = ` instead.

tests/common/runner.rs
hai-rise: We can remove `must_succeed` now because it only served mainnet blocks that are now served by `must_match_receipts_root`. The mocked Alloy blocks inheriting `must_succeed = true` is only a small artifact and their `test_execute_revm`'s counterpart has `must_succeed` hardcoded to `false` anyway.

If one day we want `must_succeed` for mocked blocks then we'll redo it properly for both mocked REVM blocks & Alloy blocks. For now the extra complexity isn't worth it.

tests/common/runner.rs
hai-rise: No, the two well-formed functions/steps can still be kept if you move the call to `build_receipt_envelopes` into `calculate_receipt_root`. The point is the caller doesn't need to know that, either in test modules or when we add it to the library's API. This is like asking PEVM user to call `execute(prepropcess(&block))` instead of calling `execute(&block)`.

tests/mainnet.rs
hai-rise: I see. We can push a step further to fallback when `RPC_URL` fails to parse:
```rust
    let rpc_url = std::env::var("RPC_URL")
        .ok()
        .and_then(|value| value.parse().ok())
        .unwrap_or(Url::parse("https://eth.llamarpc.com").unwrap());
```
But I guess it's okay to panic for the developer to double-check the value?

src/pevm.rs
```diff
@@ -207,7 +222,9 @@ pub fn execute_revm<S: Storage + Send + Sync>(
             );
             result_and_state
         })
-        .collect())
+        .collect::<Vec<_>>();
```
hai-rise, line 225: We don't need to `collect` anymore, which is also slow.

tests/common/runner.rs
```diff
@@ -50,30 +53,90 @@ pub fn test_execute_revm<S: Storage + Clone + Send + Sync>(
...
-        pevm::execute_revm_sequential(storage.clone(), spec_id, block_env.clone(), txs.clone()),
-        pevm::execute_revm(storage, spec_id, block_env, txs, concurrency_level),
+        &pevm::execute_revm_sequential(storage.clone(), spec_id, block_env.clone(), txs.clone()),
+        &pevm::execute_revm(storage, spec_id, block_env, txs, concurrency_level),
         false, // TODO: Parameterize this
     );
 }
 
+// Refer to section 4.3.2. Holistic Validity in the Ethereum Yellow Paper.
+// See @ethereum/go-ethereum/cmd/era/main.go:289.
```
hai-rise, line 63: Copy-pastable links are preferred.

tests/common/runner.rs
```diff
@@ -50,30 +53,90 @@ pub fn test_execute_revm<S: Storage + Clone + Send + Sync>(
...
+    let entries = receipt_envelopes
+        .into_iter()
+        .enumerate()
+        .map(|(index, receipt)| {
+            let key_buffer = alloy_rlp::encode_fixed_size(&index);
+            let mut value_buffer = Vec::new();
+            receipt.encode_2718(&mut value_buffer);
+            (key_buffer, value_buffer)
+        })
+        .collect::<BTreeMap<_, _>>();
```
hai-rise, line 74: This is a non-trivial type so let's annotate it fully at `let entries: Type` and just put `.collect()` here.

tests/common/runner.rs
```diff
@@ -50,30 +53,90 @@ pub fn test_execute_revm<S: Storage + Clone + Send + Sync>(
...
+
+    // We use BTreeMap because the keys must be sorted in ascending order.
```
hai-rise, line 76: This comment belongs to `entries`. We don't need to say `hash_builder.add_leaf()` would panic if that is the spec we're following.

tests/common/runner.rs
```diff
@@ -29,8 +32,8 @@ pub fn build_in_mem(
 
 // TODO: Pass in hashes to checksum, especially for real blocks.
 pub fn assert_execution_result(
-    sequential_result: PevmResult,
-    parallel_result: PevmResult,
+    sequential_result: &PevmResult,
+    parallel_result: &PevmResult,
     must_succeed: bool,
```
hai-rise, line 37: We can remove this now.

tests/common/runner.rs
```diff
@@ -50,30 +53,90 @@ pub fn test_execute_revm<S: Storage + Clone + Send + Sync>(
...
     );
+    let parallel_result = pevm::execute(
+        storage,
+        block.clone(),
+        parent_header,
+        concurrency_level,
+        false,
+    );
+    assert_execution_result(&sequential_result, &parallel_result, true);
+    let tx_results = sequential_result.unwrap();
```
hai-rise, line 130: Move this inside the `if`.

tests/ethereum/main.rs
```diff
@@ -143,29 +142,13 @@ fn run_test_unit(path: &Path, unit: &TestUnit) {
...
-                            reason: SuccessReason::Stop,
-                            ..
-                        }
-                    ));
+                    assert!(exec_results[0].receipt.status);
+                    // This is overly strict as we only need the newly created account's code to be empty.
+                    // Extracting such account is unjustified complexity so let's live with this for now.
+                    assert!(exec_results[0].state.values().all(|account| account.info.is_empty_code_hash()));
                 }
+                // Skipping special cases where REVM returns `Ok` instead on unsupported features.
```
hai-rise, line 150: My previous comment was bad... Let's remove `instead` here if possible :pray:.

tests/mainnet.rs
```diff
@@ -44,7 +44,6 @@ fn mainnet_blocks_from_rpc() {
         let rpc_storage = RpcStorage::new(provider, BlockId::number(block_number - 1));
         let db = CacheDB::new(&rpc_storage);
         common::test_execute_alloy(db.clone(), block.clone(), None, true);
-
```
hai-rise, line 47: Revert no-op!

hai-rise (approved):

Issue #138, opened 2024-06-05
Snapshot block hashes for real blocks read from RPC


hai-rise:
- [ ] Add block hashes to `RpcStorage`'s `cache`:

https://github.com/risechain/pevm/blob/d227e2611d83d94d8bf97168e7e4de1253154e62/src/storage/rpc.rs#L35-L36

https://github.com/risechain/pevm/blob/d227e2611d83d94d8bf97168e7e4de1253154e62/src/storage/rpc.rs#L133-L142

- [ ] Snapshot them to disk:

https://github.com/risechain/pevm/blob/d227e2611d83d94d8bf97168e7e4de1253154e62/tests/mainnet.rs#L48-L57

- [ ] Properly load them from disk into in-memory storage:

https://github.com/risechain/pevm/blob/d227e2611d83d94d8bf97168e7e4de1253154e62/src/storage/in_memory.rs#L24-L26

https://github.com/risechain/pevm/blob/d227e2611d83d94d8bf97168e7e4de1253154e62/src/storage/in_memory.rs#L72-L80

https://github.com/risechain/pevm/blob/d227e2611d83d94d8bf97168e7e4de1253154e62/src/storage/in_memory.rs#L83-L97

This will get the correct block hashes needed by block 19933597 to calculate the correct receipt root and unblock #134.

PR #139, opened 2024-06-06
feat: load and save block hashes

kien-rise:
Resolves https://github.com/risechain/pevm/issues/138

TODO:
- [x] Use BTreeMap when writing data
- [x] Rename `state_for_execution.json` to `pre_state.json`


tests/common/runner.rs
```diff
@@ -20,11 +19,24 @@ pub fn mock_account(idx: usize) -> (Address, PlainAccount) {
...
+// Build InMemoryStorage.
 pub fn build_in_mem(
     accounts: impl IntoIterator<Item = (Address, PlainAccount)>,
 ) -> InMemoryStorage {
-    accounts.into_iter().collect::<ChainState>().into()
+    build_in_mem_with_block_hashes(accounts, [])
+}
+
+// Build InMemoryStorage with block hashes.
+pub fn build_in_mem_with_block_hashes(
```
hai-rise, line 30: Our use cases are too small to justify this complexity (new dev asking which function to call for a new test). I'd add `, []` to the all existing call sites.

I think we can improve the existing `build_in_mem` name, too. `build_in_mem_storage` or just `build_storage`?

tests/common/mod.rs
```diff
@@ -61,9 +63,22 @@ pub fn for_each_block_from_disk(mut handler: impl FnMut(Block, ChainState)) {
 
         // Parse state
         let mut accounts: HashMap<Address, PlainAccount> = serde_json::from_reader(BufReader::new(
-            File::open(format!("blocks/{block_number}/state_for_execution.json")).unwrap(),
+            File::open(format!("blocks/{block_number}/pre_state.json")).unwrap(),
         ))
         .unwrap();
+
+        // Parse block hashes
+        let block_hashes: BlockHashes =
```
hai-rise, line 71: Not very important but the Haskell way is...

```rust
let block_hashes: BlockHashes =
    File::open(format!("blocks/{block_number}/block_hashes.json"))
        .map(|file| {
            serde_json::from_reader::<_, HashMap<U256, B256, ahash::RandomState>>(
                BufReader::new(file),
            )
            .unwrap()
            .into()
        })
        .unwrap_or_default();
```

tests/mainnet.rs
```diff
@@ -1,23 +1,27 @@
...
 pub mod common;
 
 #[test]
 fn mainnet_blocks_from_rpc() {
-    let rpc_url: Url = std::env::var("RPC_URL")
-        .unwrap_or("https://eth.llamarpc.com".to_string())
-        .parse()
-        .unwrap();
+    let rpc_url = match std::env::var("RPC_URL") {
+        Ok(value) if !value.is_empty() => value.parse().unwrap(),
```
hai-rise, line 22: I think this is non-trivial enough to add a comment:

```
The empty check is for GitHub Actions where the variable is set with an empty string when unset!?
```

tests/mainnet.rs
```diff
@@ -53,20 +57,30 @@ fn mainnet_blocks_from_rpc() {
             fs::create_dir_all(dir.clone()).unwrap();
             let file_block = File::create(format!("{dir}/block.json")).unwrap();
             serde_json::to_writer(file_block, &block).unwrap();
-            let file_state = File::create(format!("{dir}/state_for_execution.json")).unwrap();
-            serde_json::to_writer(file_state, &rpc_storage.get_cache()).unwrap();
+
+            let accounts: BTreeMap<Address, PlainAccount> =
```
hai-rise, line 61: I'd add a quick comment:
```
// We convert to `BTreeMap`s for consistent ordering & diffs
```

tests/mainnet.rs
```diff
@@ -53,20 +57,30 @@ fn mainnet_blocks_from_rpc() {
...
+                serde_json::to_writer(file, &block_hashes).unwrap();
+            }
         }
     }
 }
 
 #[test]
 fn mainnet_blocks_from_disk() {
-    common::for_each_block_from_disk(|block, state| {
+    common::for_each_block_from_disk(|block, state, block_hashes| {
```
hai-rise, line 78: If it's not much of a hassle, renaming these `state`s as `pre_state`s will make it clearer.

src/storage/rpc.rs
```diff
@@ -33,7 +33,8 @@ pub struct RpcStorage {
     // back to our `Storage` trait.
     // Not using `AHashMap` for ease of serialization.
     // TODO: Cache & snapshot block hashes too!
```
hai-rise, line 35: Can remove now!

src/storage/in_memory.rs
```diff
@@ -27,6 +36,17 @@ pub struct InMemoryStorage {
 }
 
 impl InMemoryStorage {
+    /// Create a new InMemoryStorage
+    pub fn new(
```
hai-rise, line 40: Can remove the 
```
// Currently unused. Add a popuplating
// function when needed.
```
comment above now.

hai-rise: We can also remove:

```
impl From<AHashMap<Address, PlainAccount>> for InMemoryStorage {
    fn from(accounts: AHashMap<Address, PlainAccount>) -> Self {
        let mut storage = InMemoryStorage::default();
        for (address, account) in accounts {
            storage.insert_account(
                address,
                InMemoryAccount {
                    basic: account.info.into(),
                    storage: account.storage.into_iter().collect(),
                },
            );
        }
        storage
    }
}
```

to make everyone go through `::new` now.

tests/common/runner.rs
```diff
@@ -20,11 +19,24 @@ pub fn mock_account(idx: usize) -> (Address, PlainAccount) {
...
+    accounts: impl IntoIterator<Item = (Address, PlainAccount)>,
+    block_hashes: impl IntoIterator<Item = (U256, B256)>,
+) -> InMemoryStorage {
+    InMemoryStorage::new(
+        accounts
+            .into_iter()
+            .map(|(address, account)| (address, account.into())),
```
hai-rise, line 37: This is for a future PR but the world will be a better place when we replace all `PlainAccount` with our `Account` (after renaming from `InMemoryAccount`). Then I think the only conversion needed is in the state test where we would convert our `Account` to `PlainAccount` to call their `state_merkle_trie_root` function. In a further future we'd implement that ourselves, too.

The end goal is to have consistent types both internally and for library users and minimize leaking underlying types like REVM's anywhere.

tests/mainnet.rs
hai-rise: Oh wait, it's not working as intended yet. More information: https://github.com/serde-rs/json/issues/1000#issuecomment-1464980111.

src/storage/rpc.rs
kien-rise: > // Not using `AHashMap` for ease of serialization.

Actually, since we are customizing the serialization, let's use `AHashMap`, then format the output later.

tests/common/mod.rs
kien-rise: Both the current way and the Haskell way are not looking satisfying to me. Let me try to make it cleaner.

kien-rise: I think the Haskell way is fine.

tests/common/runner.rs
kien-rise: I will change the function signature of `InMemoryStorage::new` to:

```
pub fn new(
        accounts: impl IntoIterator<Item = (Address, impl Into<InMemoryAccount>)>,
        block_hashes: impl IntoIterator<Item = (U256, B256)>,
    ) -> Self
```

tests/common/runner.rs
kien-rise: ok, I will add a comment

tests/mainnet.rs
kien-rise: let's postpone this issue for now

PR #147, opened 2024-06-11
refactor: avoid using type EvmState from revm in the pub API

kien-rise:


src/pevm.rs
```diff
@@ -44,15 +42,19 @@ pub enum PevmError {
     UnreachableError,
 }
 
+/// Represents the state transitions of the EVM accounts during execution.
```
hai-rise, line 45: `during` -> `after`.

src/pevm.rs
```diff
@@ -44,15 +42,19 @@ pub enum PevmError {
...
+/// If the value is `None`, it indicates that the account is marked as self-destructed.
+/// If the value is `Some(_)`, it indicates that the account has undergone changes.
```
hai-rise, line 47:
```
If the value is `Some(NewAccount)`, it indicates that the account has become `NewAccount`
```

src/pevm.rs
```diff
@@ -456,6 +458,11 @@ fn transform_output(
                 cumulative_gas_used,
                 logs: result.into_logs(),
             };
+            let state: EvmStateTransitions = state
+                .into_iter()
+                .filter(|(_, v)| v.is_touched())
+                .map(|(k, v)| (k, EvmAccount::from_revm_account(v)))
```
hai-rise, line 464: `k` -> `address` & `v` -> `address` for readability.

src/storage.rs
```diff
@@ -1,15 +1,61 @@
...
+    }
+
+    /// Converts a `revm::primitives::Account` into an `Option<InMemoryAccount>`.
+    /// Returns `Some(InMemoryAccount)` if the account is not self-destructed, otherwise returns `None`.
+    pub fn from_revm_account(account: revm::primitives::Account) -> Option<Self> {
+        assert!(account.is_touched());
+        if account.is_selfdestructed() {
+            None
+        } else {
+            Some(EvmAccount {
```
hai-rise, line 43: This is better expressed as a `From` instance.

src/storage.rs
```diff
@@ -1,15 +1,61 @@
...
+impl EvmAccount {
+    /// Checks if the account is empty.
+    /// An account is considered empty if its code is empty, balance is zero, and nonce is zero.
+    pub fn is_empty(&self) -> bool {
+        self.basic.code.is_empty() && self.basic.balance == U256::ZERO && self.basic.nonce == 0
+    }
+
+    /// Converts a `revm::primitives::Account` into an `Option<InMemoryAccount>`.
+    /// Returns `Some(InMemoryAccount)` if the account is not self-destructed, otherwise returns `None`.
+    pub fn from_revm_account(account: revm::primitives::Account) -> Option<Self> {
```
hai-rise, line 38: I'd inline this into `transform_output` as it's not used anywhere else and the context is non-trivial. After inlining we can remove the ad-hoc `assert`, turn the actual conversion into a `From` instance, and de-risk misusage (using this function to convert between account primitives outside of a state transition context).

src/storage.rs
```diff
@@ -1,15 +1,61 @@
...
 
+use ahash::AHashMap;
 use alloy_primitives::{Address, Bytes, B256, U256};
 use revm::{
+    db::PlainAccount,
     primitives::{AccountInfo, Bytecode},
     DatabaseRef,
 };
 
+/// An account stored in memory.
```
hai-rise, line 11: Better comment now:
```
/// An EVM account.
// TODO: Flatten `AccountBasic` or more ideally, replace this with an Alloy type.
// `AccountBasic` works for now as we're tightly tied to REVM types, hence
// conversions between `AccountBasic` & `AccountInfo` are very convenient.
```

tests/ethereum/main.rs
```diff
@@ -145,7 +145,9 @@ fn run_test_unit(path: &Path, unit: &TestUnit) {
                     assert!(exec_results[0].receipt.status.coerce_status());
                     // This is overly strict as we only need the newly created account's code to be empty.
                     // Extracting such account is unjustified complexity so let's live with this for now.
-                    assert!(exec_results[0].state.values().all(|account| account.info.is_empty_code_hash()));
+                    assert!(exec_results[0].state.values().all(|account| account
+                        .as_ref()
+                        .map_or(false, |account| account.basic.code.is_empty())));
```
hai-rise, line 150: This should be `true`? Existing accounts getting self-destructed doesn't conflict with the condition that all changed accounts must not have code attached.

If `false` is correct then `is_some_and` should read better.

src/storage.rs
```diff
@@ -1,15 +1,61 @@
...
+        EvmAccount {
+            basic: account.info.into(),
+            storage: account.storage.into_iter().collect(),
+        }
+    }
+}
+
+impl EvmAccount {
+    /// Checks if the account is empty.
+    /// An account is considered empty if its code is empty, balance is zero, and nonce is zero.
```
hai-rise, line 31: Remove as this doesn't add much to the code.

tests/ethereum/main.rs
```diff
@@ -192,23 +194,18 @@ fn run_test_unit(path: &Path, unit: &TestUnit) {
...
                     for (address, account) in state {
-                        if !account.is_touched() {
-                            continue;
-                        }
-                        if account.is_selfdestructed()
-                            || (account.is_empty()
-                                && spec_id.is_enabled_in(SpecId::SPURIOUS_DRAGON))
-                        {
+                        if let Some(account) = account {
+                            if account.is_empty() && spec_id.is_enabled_in(SpecId::SPURIOUS_DRAGON) {
```
hai-rise, line 198: Looking at this, I suspect it's better to add this check to `transform_output`'s filtering.
Then the state transition logic for library users would be so clean: Extend on `Some` and delete on `None`, no extra checks, no need to worry about eras, etc.

After that, I suppose we can promote `transform_output` to `post_process_results` now (not just about formatting but now also with EVM-specific logic).

tests/ethereum/main.rs
kien-rise: yes, `true` here makes sense

src/pevm.rs
```diff
@@ -44,15 +42,19 @@ pub enum PevmError {
     UnreachableError,
 }
 
+/// Represents the state transitions of the EVM accounts after execution.
+/// If the value is `None`, it indicates that the account is marked as self-destructed.
```
hai-rise, line 46: `as self-destructed` -> `for removal` since we now cover multiple removal logic (self-destructed and EIP-161).

src/pevm.rs
```diff
@@ -458,6 +461,19 @@ fn transform_output(
                 cumulative_gas_used,
                 logs: result.into_logs(),
             };
+            let state: EvmStateTransitions = state
+                .into_iter()
+                .filter(|(_, account)| account.is_touched())
+                .map(|(address, account)| {
+                    if account.is_selfdestructed()
+                        || account.is_empty() && spec_id.is_enabled_in(SpecId::SPURIOUS_DRAGON)
```
hai-rise, line 469: Let's link to EIP-161 since this condition is not as self-described as `is_selfdestructed`.

src/storage.rs
```diff
@@ -1,15 +1,58 @@
...
+impl From<PlainAccount> for EvmAccount {
+    fn from(account: PlainAccount) -> Self {
+        EvmAccount {
+            basic: account.info.into(),
+            storage: account.storage.into_iter().collect(),
+        }
+    }
+}
+
+impl EvmAccount {
```
hai-rise, line 32: It's minor but it's cleaner to put this `impl` right below the `struct` definition with two `From` instances following aftewards.

tests/ethereum/main.rs
```diff
@@ -145,7 +145,12 @@ fn run_test_unit(path: &Path, unit: &TestUnit) {
                     assert!(exec_results[0].receipt.status.coerce_status());
                     // This is overly strict as we only need the newly created account's code to be empty.
                     // Extracting such account is unjustified complexity so let's live with this for now.
-                    assert!(exec_results[0].state.values().all(|account| account.info.is_empty_code_hash()));
+                    assert!(exec_results[0].state.values().all(|account| {
+                        match account {
```
hai-rise, line 149: Yeah this is clearer than `map_or`.

tests/ethereum/main.rs
```diff
@@ -192,23 +197,14 @@ fn run_test_unit(path: &Path, unit: &TestUnit) {
...
                     // PEVM/REVM to meet the Ethereum specs throughout the eras.
                     for (address, account) in state {
-                        if !account.is_touched() {
-                            continue;
-                        }
-                        if account.is_selfdestructed()
-                            || (account.is_empty()
-                                && spec_id.is_enabled_in(SpecId::SPURIOUS_DRAGON))
-                        {
+                        if let Some(account) = account {
```
hai-rise, line 200: Big win for library users. Hopefully, we don't miss any information for nodes to sync correctly.

PR #151, opened 2024-06-11
chore: remove recursion from try_execute

DaniPopes:
Avoids overflowing the stack in debug mode, while it's the same on release


hai-rise (approved):
:saluting_face:

PR #156, opened 2024-06-13
feat: support Optimism

kien-rise:


kien-rise:
moved to https://github.com/risechain/pevm/pull/168

PR #168, opened 2024-06-21
feat: support optimism

kien-rise:

PR #171, opened 2024-06-24
refactor: use Chain to set chain_id

kien-rise:
These changes are made to prepare for https://github.com/risechain/pevm/pull/168.

**Before**

```
CfgEnv::default()
```

**After**

```
CfgEnv::default().with_chain_id(chain.id())
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ edition = "2021"
 # that need more security guarantees even for internally hashing
 # EVM memory locations (we do not persist these hashes).
 ahash = "0.8.11"
+alloy-chains = { version = "0.1.22", features = ["serde"] }
 alloy-primitives = { version = "0.7.5", features = ["asm-keccak"] }
 alloy-rpc-types = { git = "https://github.com/alloy-rs/alloy", rev = "a4bb5f0be3eec5c8679bdab93c1482df38ba8509" }
 bitvec = "1.0.1"
@@ -28,7 +29,6 @@ alloy-transport = { git = "https://github.com/alloy-rs/alloy", rev = "a4bb5f0be3
 alloy-transport-http = { git = "https://github.com/alloy-rs/alloy", rev = "a4bb5f0be3eec5c8679bdab93c1482df38ba8509" }
 reqwest = "0.12.4"
 tokio = { version = "1.38.0", features = ["rt-multi-thread"] }
-alloy-chains = { version = "0.1.22", features = ["serde"] }
 
 [dev-dependencies]
 alloy-consensus = { git = "https://github.com/alloy-rs/alloy", rev = "a4bb5f0be3eec5c8679bdab93c1482df38ba8509" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ alloy-transport = { git = "https://github.com/alloy-rs/alloy", rev = "a4bb5f0be3
 alloy-transport-http = { git = "https://github.com/alloy-rs/alloy", rev = "a4bb5f0be3eec5c8679bdab93c1482df38ba8509" }
 reqwest = "0.12.4"
 tokio = { version = "1.38.0", features = ["rt-multi-thread"] }
+alloy-chains = { version = "0.1.22", features = ["serde"] }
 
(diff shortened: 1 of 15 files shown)


Cargo.toml
```diff
@@ -28,6 +28,7 @@ alloy-transport = { git = "https://github.com/alloy-rs/alloy", rev = "a4bb5f0be3
 alloy-transport-http = { git = "https://github.com/alloy-rs/alloy", rev = "a4bb5f0be3eec5c8679bdab93c1482df38ba8509" }
 reqwest = "0.12.4"
 tokio = { version = "1.38.0", features = ["rt-multi-thread"] }
+alloy-chains = { version = "0.1.22", features = ["serde"] }
```
hai-rise, line 31: Let's put this above `alloy-primitives`. This section is for the RPC storage dependencies.

src/pevm.rs
```diff
@@ -60,15 +62,23 @@ pub fn execute<S: Storage + Send + Sync>(
     let Some(block_env) = get_block_env(&block.header) else {
         return Err(PevmError::MissingHeaderData);
     };
-    let Some(tx_envs) = get_tx_envs(&block.transactions) else {
+    let tx_envs: Vec<_> = block.transactions.txns().map(get_tx_env).collect();
+    if tx_envs.is_empty() {
```
hai-rise, line 66: This is a major regression:
- Empty blocks are valid.
- `MissingTransactionData` is about having at least an "invalid" transaction, not having no transactions. Perhaps we can give it a better name.
- One of our current goals is to replace `unwrap`s with grateful errors returned to users. This negates one and introduces more panics instead...

src/primitives.rs
```diff
@@ -61,46 +61,42 @@ pub(crate) fn get_block_env(header: &Header) -> Option<BlockEnv> {
...
-            blob_hashes: tx.blob_versioned_hashes.clone().unwrap_or_default(),
-            max_fee_per_blob_gas: tx.max_fee_per_blob_gas.map(U256::from),
-        })
+// TODO: Properly test this.
+pub(crate) fn get_tx_env(tx: &Transaction) -> TxEnv {
+    TxEnv {
+        caller: tx.from,
+        gas_limit: tx.gas as u64,
+        gas_price: match tx.transaction_type.unwrap() {
+            0 | 1 => U256::from(tx.gas_price.expect("Missing gasprice")),
```
hai-rise, line 70: Let's avoid panics at all costs, especially in library code that already had graceful handling. My suggestion is to make this PR focus solely on the `Chain` thing and make a new one for refactoring this function instead. Let's also introduce proper error variants like `InvalidTx(...)`  instead of panicking.

tests/common/runner.rs
```diff
@@ -29,14 +30,21 @@ pub fn assert_execution_result(sequential_result: &PevmResult, parallel_result:
 // the execution results match.
 pub fn test_execute_revm<S: Storage + Clone + Send + Sync>(
     storage: S,
+    chain: Chain,
```
hai-rise, line 33: The call sites are getting quite busy so let's do the other way around instead -- remove `chain`, `spec_id`, and `block_env` from call sites and hardcode the default values here.

We already have better tests for different specs in the Ethereum tests and real blocks via the Alloy interface.

tests/ethereum/main.rs
```diff
@@ -108,6 +109,7 @@ fn run_test_unit(path: &Path, unit: &TestUnit) {
             return;
         }
         let spec_id = spec_name.to_spec_id();
+        let chain = Chain::mainnet();
```
hai-rise, line 112: We can inline this? Can also inline `spec_id` if it's not re-used anywhere.

Cargo.toml
```diff
@@ -8,6 +8,7 @@ edition = "2021"
 # that need more security guarantees even for internally hashing
 # EVM memory locations (we do not persist these hashes).
 ahash = "0.8.11"
+alloy-chains = { version = "0.1.22", features = ["serde"] }
```
hai-rise, line 11: Missed this last time but we don't need the `serde` feature?

hai-rise (approved):
:100:

PR #172, opened 2024-06-24
perf: transactions_dependencies just need count number

jackwener:
`transactions_dependencies` is just used to determine if a transaction still has predecessor dependencies.
So, we just use a `AtomicUsize` to count its size.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -132,7 +132,7 @@ type TransactionsStatus = Vec<TxStatus>;
 // TODO: Intuitively both should share a similar data structure?
 // TODO: Consider using [SmallVec] for these `[DepsList]`
 type TransactionsDependents = Vec<Vec<TxIdx>>;
-type TransactionsDependencies = HashMap<TxIdx, Vec<TxIdx>, BuildIdentityHasher>;
+type TransactionsDependenciesNum = HashMap<TxIdx, usize, BuildIdentityHasher>;
 
 // We maintain an in-memory multi-version data structure that stores for
 // each memory location the latest value written per transaction, along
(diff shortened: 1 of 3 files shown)


hai-rise:
@jackwener This is such a sharp catch, thank you so much!

Unfortunately, the current fix deadlocks when a transaction gets invalidated several times:
- Tx 10 starts with 1 dependency on tx 2 detected during preprocessing.
- Tx 2 finishes execution, reduces tx 10's number of dependencies to 0 then schedules to execute it.
- After execution, tx 10 is invalidated with a new dependency on tx 8.
- Immediately afterwards tx 8 fails validation and needs to re-execute.
- Upon re-execution, tx 8 subtracts 1 dependency from tx 10 which now underflows to the max `usize` and skips executing it!

The easiest fix is to make `transactions_dependencies_num` `AtomicIsize` so it can go into the negatives "correctly". It's fair to assume that `pevm` should timeout and fall back to sequential before any transaction fails billions of times to underflow past the smallest negative integer.

The safer option is to also count up for dependencies detected at runtime, by adding this to `add_dependency`:
```rust
if let Some(dep_num) = self.transactions_dependencies_num.get(&tx_idx) {
    dep_num.fetch_add(1, Ordering::Release);
}
```

What do you think? It'd also be nice to turn `TransactionsDependencies` into `TransactionsDependenciesNum` with the `Vec<TxIdx>` into `usize` right from pre-processing :pray:.  

P/s: this PR inspires me to turn `TransactionsDependencies` into a `Vec<usize>` to track the live number of dependencies for each transaction at runtime. Then we can generalize `add_dependency` to take in a list of dependencies instead of eagerly adding the first blocking one. This way if we estimate a transaction to depend on 3 others then we don't need to re-execute it when only 1 of the 3 finish execution to only fail validation again when the other 2 finish! It's a big change so perhaps a new PR some day :).

hai-rise (approved):
This is a huge win, thank you :tada:!

jackwener:
@hai-rise Thank you for your patient explanation and help ❤️

PR #174, opened 2024-06-24
Rename plain account evm account

### LINKED ISSUE #143: Replace `PlainAccount` with our own `Account` type
We can rename `InMemoryAccount` to `Account` and replace most `PlainAccount` usage with it. Then the only conversion needed is probably in the state test where we would convert our `Account` to `PlainAccount` to call REVM's `state_merkle_trie_root function`. In the future we can implement that ourselves, too.

The end goal is to have consistent types both internally and for library users and minimize leaking underlying types like REVM's anywhere.

chirag-bgh:
Closes #143 

- Use `EvmAccount` instead of `PlainAccount`
- Snapshot generated for all mainnet blocks
- Serialise the new `EvmAccount` types as json


chirag-bgh:
The snapshot are generated using this method:
```
if std::env::var("SNAPSHOT_BLOCKS") == Ok("1".to_string()) {
            let dir = format!("blocks/{block_number}");
            fs::create_dir_all(dir.clone()).unwrap();
            let file_block = File::create(format!("{dir}/block.json")).unwrap();
            serde_json::to_writer(file_block, &block).unwrap();

            // TODO: Snapshot with consistent ordering for ease of diffing.
            let accounts: BTreeMap<Address, EvmAccount> =
                rpc_storage.get_cache_accounts().into_iter().collect();
            let file_state = File::create(format!("{dir}/pre_state.json")).unwrap();
            serde_json::to_writer(file_state, &accounts).unwrap();

            // We convert to `BTreeMap`s for consistent ordering & diffs between snapshots
            let block_hashes: BTreeMap<U256, B256> =
                rpc_storage.get_cache_block_hashes().into_iter().collect();
            if !block_hashes.is_empty() {
                let file = File::create(format!("{dir}/block_hashes.json")).unwrap();
                serde_json::to_writer(file, &block_hashes).unwrap();
            }
        }

**Comment by kien-rise** [comment]
@chirag-bgh  CI is failing. For some reason, a `code_hash` is `Some(_)` while it should be `None`:

```
_.unwrap().0.state.get("0x95222290dd7278aa3ddd389cc1e1d165cc4bafe5").unwrap().unwrap().basic.code_hash
   left = "None"
  right = "Some(_)"
```

`None` is correct because I checked on `main` branch.

**Comment by kien-rise** [comment]
Let's make an implicit rule: If `code == None`, then `code_hash = None`.

Maybe we can add a comment in the struct definition.

**Comment by hai-rise** [comment]
@chirag-bgh It will be easier if we break this PR down into smaller ones. 

For the first one, let's just convert the in-memory accounts and keep everything on disk as `PlainAccount`. 

Then we can focus on serializing `EvmAccount` and better structuring snapshots in the next PRs. For instance, to move code-hash-to-bytecode mappings to a dedicated file and only keep code hashes in serialized accounts. Something like this:
```
/pevm/data$ tree
.
├── block_hashes.json
├── blocks
│   ├── 10760440
│   │   ├── block.json
│   │   └── pre_state.json
    ...
│   └── 930196
│       ├── block.json
│       └── pre_state.json
└── bytecodes.json
```

This would deduplicate much code from the same contract (forks) across blocks and make each `pre_state.json` more readable.
 
What do you think :pray:?

chirag-bgh:
This sounds perfect. Snapshot generation and serialization step should be done in next PRs.

PR #176, opened 2024-06-25
refactor: replace get_tx_envs by get_tx_env

kien-rise:


src/pevm.rs
```diff
@@ -62,9 +64,14 @@ pub fn execute<S: Storage + Send + Sync>(
...
+        .transactions
+        .as_transactions()
+        .ok_or(PevmError::MissingTransactionData)?
+        .iter()
+        .map(get_tx_env)
+        .collect::<Result<Vec<TxEnv>, TransactionParsingError>>()
+        .map_err(PevmError::InvalidTransaction)?;
 
     // TODO: Continue to fine tune this condition.
     if force_sequential || tx_envs.len() < 4 || block.header.gas_used <= 650_000 {
```
hai-rise, line 77: I just realized we're only using `block.header.gas_used`  after all the conversions.

Let's copy `gas_used` to a local variable so `get_tx_env` can take ownership of the `Transaction`s and avoid the expensive clones (`tx.input`, `tx.access_list`, implicit byte copies, etc.).

Currently, we're cloning the transaction data just to drop them with `block` at the end of this scope :cry:.

src/primitives.rs
```diff
@@ -58,49 +58,59 @@ pub(crate) fn get_block_env(header: &Header) -> Option<BlockEnv> {
...
-                    )
-                })
-                .collect(),
-            blob_hashes: tx.blob_versioned_hashes.clone().unwrap_or_default(),
-            max_fee_per_blob_gas: tx.max_fee_per_blob_gas.map(U256::from),
-        })
-    }
-    Some(tx_envs)
+// TODO: Properly test this.
+pub(crate) fn get_tx_env(tx: &Transaction) -> Result<TxEnv, TransactionParsingError> {
```
hai-rise, line 73: Let's take ownership now!

src/primitives.rs
```diff
@@ -58,49 +58,59 @@ pub(crate) fn get_block_env(header: &Header) -> Option<BlockEnv> {
...
+        gas_price: match tx.transaction_type.unwrap() {
+            0 | 1 => U256::from(
+                tx.gas_price
+                    .ok_or(TransactionParsingError::MissingGasPrice)?,
+            ),
+            2 | 3 => U256::from(
+                tx.max_fee_per_gas
+                    .ok_or(TransactionParsingError::MissingMaxFeePerGas)?,
+            ),
+            unknown_value => return Err(TransactionParsingError::InvalidType(unknown_value)),
```
hai-rise, line 86: `unknown` is more concise. And yes, returning an error here is better than the previous version of optimistically predicting compatibility.

src/primitives.rs
```diff
@@ -58,49 +58,59 @@ pub(crate) fn get_block_env(header: &Header) -> Option<BlockEnv> {
...
-        })
-    }
-    Some(tx_envs)
+// TODO: Properly test this.
+pub(crate) fn get_tx_env(tx: &Transaction) -> Result<TxEnv, TransactionParsingError> {
+    Ok(TxEnv {
+        caller: tx.from,
+        gas_limit: tx.gas as u64,
+        gas_price: match tx.transaction_type.unwrap() {
+            0 | 1 => U256::from(
```
hai-rise, line 78: We can wrap just one `U256::from` around the `match`?

src/primitives.rs
```diff
@@ -58,49 +58,59 @@ pub(crate) fn get_block_env(header: &Header) -> Option<BlockEnv> {
...
-            blob_hashes: tx.blob_versioned_hashes.clone().unwrap_or_default(),
-            max_fee_per_blob_gas: tx.max_fee_per_blob_gas.map(U256::from),
-        })
-    }
-    Some(tx_envs)
+// TODO: Properly test this.
+pub(crate) fn get_tx_env(tx: &Transaction) -> Result<TxEnv, TransactionParsingError> {
+    Ok(TxEnv {
+        caller: tx.from,
+        gas_limit: tx.gas as u64,
```
hai-rise, line 76: Let's keep the previous `try_into`  and give this a specific error. Else this would panic on users.

src/pevm.rs
hai-rise: I just tried this locally and it turns out we don't need to copy `gas_used`. Rust's smart enough to know the partial move of the block transactions doesn't clash with its header. Nice!

src/primitives.rs
kien-rise: I recommend not doing so. Currently, it is only possible because `tx.gas_price` has the same type as `tx.max_fee_per_gas` (`Option<u128>`).

Later if we support optimism, the added arm will be:

```
#[cfg(feature = "optimism")]
126 if chain.is_optimism() => U256::ZERO,
```

src/primitives.rs
kien-rise: fyi, `as u64` will truncate high bits and not panic. But we can use `try_into` if we think the extra safety is needed.

src/primitives.rs
hai-rise: We can put `0` in that arm! Jk this is very minor so yeah, let's keep it as is.

PR #177, opened 2024-06-25
Use EvmAccount instead of PlainAccount

chirag-bgh:
Note: This change is not implemented for accounts on the disk. Accounts are deserialised into `PlainAccount` for now.
Snapshot generation and serialisation of `EvmAccount` types must be done in subsequent PRs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ edition = "2021"
 # Put this behind a feature flag if there are use cases & users
 # that need more security guarantees even for internally hashing
 # EVM memory locations (we do not persist these hashes).
-ahash = { version = "0.8.11", features = ["serde"] }
+ahash = { version = "0.8.11" }
 alloy-chains = { version = "0.1.22" }
 alloy-primitives = { version = "0.7.5", features = ["asm-keccak"] }
 alloy-rpc-types = { git = "https://github.com/alloy-rs/alloy", rev = "a4bb5f0be3eec5c8679bdab93c1482df38ba8509" }
@@ -15,7 +15,6 @@ bitvec = "1.0.1"
 crossbeam = "0.8.4"
 dashmap = "6.0.0"
 defer-drop = "1.3.0"
-serde = "1.0.203"
 
 # Let's do our best to port needed REVM changes upstream
 revm = { git = "https://github.com/risechain/revm", rev = "979d069f0c2798f416c57f82ca1ebef46d257c4e", features = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ edition = "2021"
 # Put this behind a feature flag if there are use cases & users
 # that need more security guarantees even for internally hashing
 # EVM memory locations (we do not persist these hashes).
-ahash = "0.8.11"
+ahash = { version = "0.8.11", features = ["serde"] }
 alloy-chains = { version = "0.1.22" }
 alloy-primitives = { version = "0.7.5", features = ["asm-keccak"] }
 alloy-rpc-types = { git = "https://github.com/alloy-rs/alloy", rev = "a4bb5f0be3eec5c8679bdab93c1482df38ba8509" }
@@ -15,6 +15,7 @@ bitvec = "1.0.1"
 crossbeam = "0.8.4"
 dashmap = "6.0.0"
 defer-drop = "1.3.0"
+serde = "1.0.203"
 
 # Let's do our best to port needed REVM changes upstream
 revm = { git = "https://github.com/risechain/revm", rev = "979d069f0c2798f416c57f82ca1ebef46d257c4e", features = [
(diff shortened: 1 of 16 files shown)


Cargo.toml
```diff
@@ -7,14 +7,15 @@ edition = "2021"
 # Put this behind a feature flag if there are use cases & users
 # that need more security guarantees even for internally hashing
 # EVM memory locations (we do not persist these hashes).
-ahash = "0.8.11"
+ahash = { version = "0.8.11", features = ["serde"] }
```
hai-rise, line 10: Can we save the `Cargo.*` changes to the next PR for better traceability?

src/storage.rs
```diff
@@ -9,19 +9,32 @@ use revm::{
     primitives::{Account, AccountInfo, Bytecode, JumpTable},
     DatabaseRef,
 };
+use serde::{Deserialize, Serialize};
```
hai-rise, line 12: Can we save the `Cargo.*` changes to the next PR for better traceability?

src/storage.rs
```diff
@@ -9,19 +9,32 @@ use revm::{
...
     pub basic: AccountBasic,
     /// The account's storage.
     pub storage: AHashMap<U256, U256>,
 }
 
+impl EvmAccount {
+    /// Create `PlainAccount` from `EvmAccount`.
+    ///
+    /// This is useful for testing and conversion between storage types.
+    pub fn into_plain(self) -> PlainAccount {
```
hai-rise, line 30: This will be cleaner as an `impl From<EvmAccount> for PlainAccount`.

src/storage.rs
```diff
@@ -59,11 +81,29 @@ pub struct AccountBasic {
 }
 
 impl AccountBasic {
+    /// Create a new account with the given balance, nonce, code hash and code.
```
hai-rise, line 84: The `struct` is pretty simple with all `pub` fields so constructing with `AccountBasic {` should be good enough for now.

src/storage.rs
```diff
@@ -59,11 +81,29 @@ pub struct AccountBasic {
...
+    }
+
     /// Check if an account is empty for removal per EIP-161
     // https://github.com/ethereum/EIPs/blob/96523ef4d76ca440f73f0403ddb5c9cb3b24dcae/EIPS/eip-161.md
     pub fn is_empty(&self) -> bool {
         self.balance == U256::ZERO && self.nonce == 0 && self.code.is_none()
     }
+
+    /// Create a new account with the given balance.
+    pub fn from_balance(balance: U256) -> Self {
```
hai-rise, line 101: It looks like all use cases combine this with `impl From<AccountBasic> for EvmAccount`. 

I wonder if we can simplify both here and at call sites by providing `EvmAccount::with_balance` instead?

src/storage/rpc.rs
```diff
@@ -107,10 +108,11 @@ impl DatabaseRef for RpcStorage {
             }
             let code = Bytecode::new_raw(code);
             let info = AccountInfo::new(balance, nonce, code.hash_slow(), code);
```
hai-rise, line 110: Ah, it looks like the cleaner path is to construct `AccountBasic` & `EvmAccount` directly, instead of going through REVM types before converting.

tests/erc20/contract.rs
```diff
@@ -87,9 +84,14 @@ impl ERC20Token {
...
-            info: AccountInfo::new(U256::ZERO, 1u64, bytecode.hash_slow(), bytecode.clone()),
-            storage: store.build(),
+        EvmAccount {
+            basic: AccountBasic::new(
+                U256::ZERO,
+                1u64,
+                bytecode.hash_slow(),
+                bytecode.clone().into(),
+            ),
+            storage: store.build().into_iter().collect(),
```
hai-rise, line 94: Maybe it's a good time to make `store.build()` returns `AHashMap` now!

tests/mainnet.rs
```diff
@@ -65,8 +65,11 @@ fn mainnet_blocks_from_rpc() {
 
             // TODO: Snapshot with consistent ordering for ease of diffing.
             // Currently PlainAccount's storage ordering isn't consistent.
-            let accounts: BTreeMap<Address, PlainAccount> =
-                rpc_storage.get_cache_accounts().into_iter().collect();
+            let accounts: BTreeMap<Address, PlainAccount> = rpc_storage
+                .get_cache_accounts()
+                .into_iter()
```
hai-rise, line 70: Now I see why you were motivated to store `EvmAccount` directly on disk in the same PR :pray:. 

Let's do it next!

hai-rise (changes requested):
Almost there! Just a few nit comments left for consistency.

Thank you again for your time :saluting_face:.

hai-rise (approved):
🫡 🔥

Issue #60, opened 2024-05-13
Support custom chains


hai-rise:
For instance, to have custom chain specs with different hardfork times, OP support, etc.

PR #179, opened 2024-06-25
feat: generalize reward policy to support other chains

kien-rise:
One step towards https://github.com/risechain/pevm/issues/60.

### DIFF
--- a/src/vm.rs
+++ b/src/vm.rs
@@ -1,5 +1,3 @@
-use std::cmp::min;
-
 use ahash::AHashMap;
 use alloy_chains::Chain;
 use alloy_rpc_types::Receipt;
@@ -7,9 +5,7 @@ use defer_drop::DeferDrop;
 use revm::{
     primitives::{
         AccountInfo, Address, BlockEnv, Bytecode, CfgEnv, EVMError, Env, InvalidTransaction,
-        ResultAndState,
-        SpecId::{self, LONDON},
-        TransactTo, TxEnv, B256, U256,
+        ResultAndState, SpecId, TransactTo, TxEnv, B256, U256,
     },
     Context, Database, Evm, EvmContext, Handler,
 };
@@ -27,6 +23,10 @@ pub type ExecutionError = EVMError<ReadError>;
 /// If the value is [None], it indicates that the account is marked for removal.
 /// If the value is [Some(new_state)], it indicates that the account has become [new_state].
 type EvmStateTransitions = AHashMap<Address, Option<EvmAccount>>;
+
+pub(crate) enum RewardPolicy {
+    Ethereum,
+}
 
 /// Execution result of a transaction
 #[derive(Debug, Clone, PartialEq)]
@@ -359,6 +359,7 @@ pub(crate) struct Vm<'a, S: Storage> {
     spec_id: SpecId,
     block_env: BlockEnv,
     beneficiary_location_hash: MemoryLocationHash,
+    reward_policy: &'a RewardPolicy,
     // TODO: Make REVM [Evm] or at least [Handle] thread safe to consume
     // the [TxEnv] into them here, to avoid heavy re-initialization when
     // re-executing a transaction.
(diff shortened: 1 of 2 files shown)


src/vm.rs
```diff
@@ -27,6 +24,10 @@ pub type ExecutionError = EVMError<ReadError>;
 /// If the value is [Some(new_state)], it indicates that the account has become [new_state].
 type EvmStateTransitions = AHashMap<Address, Option<EvmAccount>>;
 
+pub(crate) enum RewardPolicy {
```
hai-rise, line 27: We can keep this private to `Vm` for now.

src/vm.rs
```diff
@@ -376,6 +378,7 @@ impl<'a, S: Storage> Vm<'a, S> {
             spec_id,
             beneficiary_location_hash: hasher.hash_one(MemoryLocation::Basic(block_env.coinbase)),
             block_env,
+            reward_policy: &RewardPolicy::Ethereum, // derived from chain
```
hai-rise, line 381: `TODO: Derive from [chain]`

src/vm.rs
```diff
@@ -550,6 +532,35 @@ impl<'a, S: Storage> Vm<'a, S> {
             Err(err) => VmExecutionResult::ExecutionError(err),
         }
     }
+
+    // Apply rewards (balance increments) based on self.reward_policy
```
hai-rise, line 536:
```
Apply rewards (balance increments) to beneficiary accounts, etc.
```

We also base on `block_env` and `tx` so `based on self.reward_policy` is not very precise.
Also, `reward_policy` is more of a cache to avoid recalculating block-specific values, not everything needed to calculate the rewards.

src/vm.rs
```diff
@@ -550,6 +532,35 @@ impl<'a, S: Storage> Vm<'a, S> {
...
+                    gas_price = gas_price.saturating_sub(self.block_env.basefee);
+                }
+                vec![(self.beneficiary_location_hash, gas_price * gas_used)]
+            }
+        };
+
+        for (recipient, amount) in rewards {
+            if let Some((_, value)) = write_set.iter_mut().find(|entry| entry.0 == recipient) {
+                match value {
+                    MemoryValue::Basic(account_info) => account_info.balance += amount,
```
hai-rise, line 555: For more consistency with the rest of the code base:
- Use `info` where there is a conversion with `AccountBasic` nearby.
- Use `account` otherwise as the default name.

src/vm.rs
hai-rise: Should also add a comment like:

```
// Different chains may have different reward policies.
// This enum indicates which policy to follow with optional
// data to aid reward calculations. 
```

src/vm.rs
```diff
@@ -550,6 +532,35 @@ impl<'a, S: Storage> Vm<'a, S> {
...
+                };
+                if self.spec_id.is_enabled_in(SpecId::LONDON) {
+                    gas_price = gas_price.saturating_sub(self.block_env.basefee);
+                }
+                vec![(self.beneficiary_location_hash, gas_price * gas_used)]
+            }
+        };
+
+        for (recipient, amount) in rewards {
+            if let Some((_, value)) = write_set.iter_mut().find(|entry| entry.0 == recipient) {
```
hai-rise, line 553: `find(|(location, _)| location == recipient)`?

src/vm.rs
```diff
@@ -550,6 +532,35 @@ impl<'a, S: Storage> Vm<'a, S> {
...
+                match value {
+                    MemoryValue::Basic(account_info) => account_info.balance += amount,
+                    MemoryValue::LazyBalanceAddition(addition) => *addition += amount,
+                    MemoryValue::Storage(_) => unreachable!(),
```
hai-rise, line 557: We can return `InvalidMemoryLocationType` here. Perhaps `UnexpectedMemoryLocation` is a better name.

hai-rise (approved):

PR #180, opened 2024-06-25
Implement `Storage` for `RpcStorage` instead of `Database`

hai-rise:


src/storage/rpc.rs
```diff
@@ -64,29 +65,25 @@ impl RpcStorage {
...
-// Going with Revm's [Database] simply to make it easier
-// to try matching sequential & parallel execution.
-// In the future we should match block roots anyway.
-impl DatabaseRef for RpcStorage {
+impl Storage for RpcStorage {
     type Error = TransportError;
 
-    fn basic_ref(&self, address: Address) -> Result<Option<AccountInfo>, Self::Error> {
-        if let Some(account) = self.cache_accounts.lock().unwrap().get(&address) {
-            return Ok(Some(account.basic.clone().into()));
```
hai-rise, line 76: cc @chirag-bgh I could remove most conversions between pevm & revm types after paying another tech debt now. Changing the cache to `EvmAccount` triggered this, thanks!

PR #184, opened 2024-06-27
chore: convert TransactionsDependenciesNum into Vec

jackwener:
convert TransactionsDependenciesNum into Vec from HashMap.

Remove part of TODO above TransactionsDependenciesNum.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -128,7 +128,7 @@ struct TxStatus {
 type TransactionsStatus = Vec<TxStatus>;
 // TODO: Consider using [SmallVec] for these `[DepsList]`
 type TransactionsDependents = Vec<Vec<TxIdx>>;
-type TransactionsDependenciesNum = HashMap<TxIdx, usize, BuildIdentityHasher>;
+type TransactionsDependenciesNum = Vec<usize>;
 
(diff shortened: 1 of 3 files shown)


hai-rise:
@jackwener Thank you again for the effort!

I'd keep this open for further experiments, as it regresses big blocks with few preprocessed dependencies but many runtime dependencies:
- A `Vec` of guaranteed `block_size` is more expensive than a mostly empty `HashMap`.
- There are two extra `Atomic` operations per detected runtime dependency (even for transactions without preprocessed transactions).

This may be justified when we track multiple runtime dependencies per transaction. My first attempt at 
```
// TODO: we can make [add_dependency] take in a list instead of just the first estimated one.
```
was a regression though :cry:.

jackwener:
Thank you for sharing these idea and result of test

hai-rise:
We have removed preprocessing as lazy updates have shown superior performance (for less code). 
We therefore don't track the number of dependencies per transaction in the scheduler anymore, and I have closed the related issue: https://github.com/risechain/pevm/issues/195#issuecomment-2214599520.
Perhaps we can retire this PR now 🙏 ?

hai-rise:
@jackwener Retagging in case you've missed the previous message :pray:

PR #186, opened 2024-06-27
chore: use match instead of double if

jackwener:
use `match` instead of double `if`

It will make code more clear.

### DIFF
--- a/src/vm.rs
+++ b/src/vm.rs
@@ -266,52 +266,53 @@ impl<'a, S: Storage> Database for VmDb<'a, S> {
             }
         }
 
-        // Fall back to storage
-        if final_account.is_none() {
-            // Populate [Storage] on the first read
-            if !has_prev_origins {
-                new_origins.push(ReadOrigin::Storage);
-            }
-            // Inconsistent: previous origin is longer or didn't read
-            // from storage for the last origin.
-            else if read_origins.len() != new_origins.len() + 1
-                || read_origins.last() != Some(&ReadOrigin::Storage)
-            {
-                return Err(ReadError::InconsistentRead);
-            }
-            final_account = match self.vm.storage.basic(&address) {
-                Ok(Some(account)) => {
-                    let mut info = AccountInfo::from(account);
-                    info.balance += balance_addition;
-                    Some(info)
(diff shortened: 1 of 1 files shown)


hai-rise:
Intuitively, if we cannot find `final_account` in `MvMemory`, we would fetch it from `Storage` before registering the **final** value to the read cache. This change seems to not write to the cache if the value is read from storage. I'm surprised that this passes the tests. That could be an optimization we can look into!

Also, I'm hesitant to move the straightforward cache logic into a rather complicated value evaluation & origin tracking code. So yeah, let's keep this open until I can find the time to look into the skip-cache optimization 🙏.

jackwener:
> this passes the tests. 

I feel like you may have some misunderstandings about my code. In my opinion, this PR does not change the code logic.

In this PR, I just move some code into `match`. You can see https://github.com/risechain/pevm/pull/186/files?diff=split&w=1 , this diff is more clear(ignore space_change)

hai-rise:
There is a semantic change as `final_account` can still be set via `match self.vm.storage.basic(&address) {` in the previous `if final_account.is_none() {` branch, right?

So the previous behaviour is that: find `final_account` from `MvMemory`, if not found then try finding it from `Storage`, then write it to `self.read_set.accounts` if we could find one.

The new behaviour here is that it only writes to `self.read_set.accounts` if `final_account` is found from `MvMemory`, and doesn't write if it has fallen back to `Storage.

jackwener:
> There is a semantic change as final_account can still be set via match self.vm.storage.basic(&address) { in the previous if final_account.is_none() {

Get it, Thank you !

hai-rise:
> Hahaha, I feel the same way

Reading accounts from storage is a hot path so it would be a huge win if we could remove the (potentially unneeded) cache write. I'll have a look 🫡.

hai-rise:
@jackwener So the tests still passed because not registering read accounts would make all written values "new" and be recorded in the multi-version data structure. This would hurt performance a lot as ideally we would only record changed values, which we need the read accounts cache to distinguish.

The related code has also changed quite a lot, so I'd suggest closing this PR now 🙏.

PR #198, opened 2024-07-01
feat: unify logic to handle rewarded accounts (beneficiary account and lazy recipients)

kien-rise:
one step toward https://github.com/risechain/pevm/issues/60

### DIFF
--- a/src/pevm.rs
+++ b/src/pevm.rs
@@ -239,7 +239,8 @@ pub fn execute_revm<S: Storage + Send + Sync>(
                     account.basic.balance = current_account.balance;
                     account.basic.nonce = current_account.nonce;
                 } else {
-                    // Implicit write: e.g. gas payments to the beneficiary account, which doesn't have explicit writes in [result.state]
+                    // Implicit write: e.g. gas payments to the beneficiary account,
+                    // which doesn't have explicit writes in [tx_result.state]
                     *account = Some(EvmAccount {
                         basic: current_account.clone(),
                         storage: AHashMap::default(),
--- a/src/pevm.rs
+++ b/src/pevm.rs
@@ -1,11 +1,13 @@
 use std::{
-    collections::{BTreeMap, HashMap, HashSet},
+    collections::{HashMap, HashSet},
     fmt::Debug,
+    iter::once,
     num::NonZeroUsize,
     sync::{Mutex, OnceLock},
     thread,
 };
 
+use ahash::AHashMap;
 use alloy_chains::Chain;
 use alloy_primitives::{Address, U256};
 use alloy_rpc_types::{Block, BlockTransactions};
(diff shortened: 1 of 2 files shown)


src/storage/rpc.rs
```diff
@@ -106,8 +106,8 @@ impl Storage for RpcStorage {
             let basic = AccountBasic {
                 balance,
                 nonce,
-                code_hash: Some(code.hash_slow()),
-                code: Some(code.into()),
+                code_hash: (!code.is_empty()).then(|| code.hash_slow()),
```
hai-rise, line 109: This isn't very clean, but I also want to move `code` out of `AccountBasic` so let's go with it for now.

src/pevm.rs
```diff
@@ -328,6 +262,54 @@ pub fn execute_revm_sequential<S: Storage>(
     Ok(results)
 }
 
+fn canonize_account(
```
hai-rise, line 265: `fully_evaluate_account` is a better name as it pairs well with lazy update (evaluation). `canonize` risks mistaking this with the concept of canonizing chains in consensus.

src/pevm.rs
```diff
@@ -328,6 +262,54 @@ pub fn execute_revm_sequential<S: Storage>(
...
+    results: &mut [PevmTxExecutionResult],
+    address: Address,
+    initial_account_basic: AccountBasic,
```
hai-rise, line 268: We can pass in `storage` and calculate `current_account` inside here with the same formula for everyone?

src/pevm.rs
```diff
@@ -328,6 +262,54 @@ pub fn execute_revm_sequential<S: Storage>(
...
+    write_history: BTreeMap<TxIdx, MemoryEntry>,
+) {
+    let mut current_account = initial_account_basic;
+
+    for (tx_idx, memory_entry) in write_history {
+        match memory_entry {
+            MemoryEntry::Data(_, MemoryValue::Basic(info)) => {
```
hai-rise, line 275: Let's not swallow comments like `// TODO: Can the code change mid-block?`.

src/pevm.rs
```diff
@@ -328,6 +262,54 @@ pub fn execute_revm_sequential<S: Storage>(
...
+                current_account.balance = info.balance;
+                current_account.nonce = info.nonce;
+            }
+            MemoryEntry::Data(_, MemoryValue::LazyBalanceAddition(addition)) => {
+                current_account.balance += addition;
+            }
+            _ => unreachable!(),
```
hai-rise, line 282: Similarly, let's put a `// TODO: Better error handling` here.

src/pevm.rs
```diff
@@ -328,6 +262,54 @@ pub fn execute_revm_sequential<S: Storage>(
...
+        }
+
+        let result = unsafe { results.get_unchecked_mut(tx_idx) };
+        match result.state.entry(address) {
+            std::collections::hash_map::Entry::Occupied(occupied) => {
```
hai-rise, line 287: Only qualify until `Entry::Occupied` for consistency with the rest of the code base.

src/pevm.rs
```diff
@@ -328,6 +262,54 @@ pub fn execute_revm_sequential<S: Storage>(
...
+                    account.basic.nonce = current_account.nonce;
+                } else {
+                    // for beneficiary only
+                    *account = Some(EvmAccount {
+                        basic: current_account.clone(),
+                        storage: Default::default(),
+                    });
+                }
+            }
+            std::collections::hash_map::Entry::Vacant(vacant) => {
```
hai-rise, line 302: Only qualify until `Entry::Occupied` for consistency with the rest of the code base.

src/pevm.rs
```diff
@@ -328,6 +262,54 @@ pub fn execute_revm_sequential<S: Storage>(
...
+                current_account.balance = info.balance;
+                current_account.nonce = info.nonce;
+            }
+            MemoryEntry::Data(_, MemoryValue::LazyBalanceAddition(addition)) => {
+                current_account.balance += addition;
+            }
+            _ => unreachable!(),
+        }
+
+        let result = unsafe { results.get_unchecked_mut(tx_idx) };
```
hai-rise, line 285: Do not swallow comments: 
```
// SAFETY: The multi-version data structure should not leak an index
// over block size.
```

src/pevm.rs
```diff
@@ -328,6 +262,54 @@ pub fn execute_revm_sequential<S: Storage>(
...
+                    // for beneficiary only
+                    *account = Some(EvmAccount {
+                        basic: current_account.clone(),
+                        storage: Default::default(),
+                    });
+                }
+            }
+            std::collections::hash_map::Entry::Vacant(vacant) => {
+                // for beneficiary only
+                vacant.insert((!current_account.is_empty()).then(|| EvmAccount {
```
hai-rise, line 304: This is only easier to read as a one-liner. For multiple lines let's just `if-else`.

src/pevm.rs
```diff
@@ -328,6 +262,54 @@ pub fn execute_revm_sequential<S: Storage>(
...
+        match result.state.entry(address) {
+            std::collections::hash_map::Entry::Occupied(occupied) => {
+                let account = occupied.into_mut();
+                if current_account.is_empty() {
+                    *account = None;
+                } else if let Some(account) = account {
+                    account.basic.balance = current_account.balance;
+                    account.basic.nonce = current_account.nonce;
+                } else {
+                    // for beneficiary only
```
hai-rise, line 295: This will be for reward addresses like OP's as well, right? Let's do 

```
For implicit rewards like the beneficiary account that doesn't have explicit writes in [result.state]
```

src/pevm.rs
```diff
@@ -328,6 +262,54 @@ pub fn execute_revm_sequential<S: Storage>(
...
+                    *account = Some(EvmAccount {
+                        basic: current_account.clone(),
+                        storage: Default::default(),
+                    });
+                }
+            }
+            std::collections::hash_map::Entry::Vacant(vacant) => {
+                // for beneficiary only
```
hai-rise, line 303: This will be for reward addresses like OP's as well, right? Let's do 

```
For implicit rewards like the beneficiary account that doesn't have explicit writes in [result.state]
```

src/pevm.rs
```diff
@@ -328,6 +262,54 @@ pub fn execute_revm_sequential<S: Storage>(
...
+        let result = unsafe { results.get_unchecked_mut(tx_idx) };
+        match result.state.entry(address) {
+            std::collections::hash_map::Entry::Occupied(occupied) => {
+                let account = occupied.into_mut();
+                if current_account.is_empty() {
+                    *account = None;
+                } else if let Some(account) = account {
+                    account.basic.balance = current_account.balance;
+                    account.basic.nonce = current_account.nonce;
+                } else {
```
hai-rise, line 294: We can remove this `else`? The beneficiary handling falls in the `Vacant` case?

src/pevm.rs
```diff
@@ -328,6 +262,54 @@ pub fn execute_revm_sequential<S: Storage>(
...
+            MemoryEntry::Data(_, MemoryValue::LazyBalanceAddition(addition)) => {
+                current_account.balance += addition;
+            }
+            _ => unreachable!(),
+        }
+
+        let result = unsafe { results.get_unchecked_mut(tx_idx) };
+        match result.state.entry(address) {
+            std::collections::hash_map::Entry::Occupied(occupied) => {
+                let account = occupied.into_mut();
```
hai-rise, line 288: `result.state.entry(address).or_default()` is much better if we always expect an output value. Like we can replace this whole `match` with:
```rust
        let account = result.state.entry(address).or_default();
        if current_account.is_empty() {
            *account = None;
        } else if let Some(account) = account {
            account.basic.balance = current_account.balance;
            account.basic.nonce = current_account.nonce;
        } else {
            *account = Some(EvmAccount {
                basic: current_account.clone(),
                storage: Default::default(),
            });
        }
```
Plus comments of course.

src/pevm.rs
hai-rise: Also nit but `tx_result` & `tx_results` would read much smoother.

src/pevm.rs
```diff
@@ -328,6 +262,54 @@ pub fn execute_revm_sequential<S: Storage>(
     Ok(results)
 }
 
+fn canonize_account(
+    results: &mut [PevmTxExecutionResult],
+    address: Address,
```
hai-rise, line 267: For performance, it's better to pass in a list of addresses (`lazy_addresses = [beneficiary_address] <> lazy_tos <> lazy_senders <> ...`) here to call this function only once. That way, the compiler can confidently inline this function and remove indirections.

hai-rise: This function is very small after refactoring so I wouldn't mind inlining it, to be frank.

src/pevm.rs
kien-rise: I don't think so.

- This `else` branch is reached when `result.state` _contains_ `address` as a key, and the corresponding value is `None`.

- The `Entry::Vacant(vacant) => {` branch is reached when `result.state` _does not contains_ `address`.

src/pevm.rs
kien-rise: This simplification makes a lot of sense. Thanks!

src/pevm.rs
hai-rise: No, the point is that the beneficiary handling case always falls on the `Vacant` case. Can the beneficiary account's nonce be destructed (getting nonce reset even) mid-block to fall into this `else` branch?

kien-rise: noted that

src/pevm.rs
kien-rise: I refactored. Pls review again, tks.

PR #199, opened 2024-07-01
chore: typo fix

jackwener:
- typo fix
- use `while let` to make code clear


src/vm.rs
```diff
@@ -243,17 +243,11 @@ impl<'a, S: Storage> Database for VmDb<'a, S> {
...
                                 }
                                 _ => return Err(ReadError::InvalidMemoryLocationType),
                             }
                         }
-                        _ => {
-                            if need_consecutive_idxs && current_idx > &0 {
-                                return reschedule;
-                            }
-                            break;
-                        }
```
jackwener, line 256: It's unreachable()

PR #206, opened 2024-07-05
refactor: extract fn encode_receipt_2718

kien-rise:

### DIFF
--- a/tests/common/runner.rs
+++ b/tests/common/runner.rs
@@ -1,6 +1,6 @@
 use alloy_chains::Chain;
 use alloy_consensus::{ReceiptEnvelope, TxType};
-use alloy_primitives::{Bloom, B256};
+use alloy_primitives::{Bloom, Bytes, B256};
 use alloy_provider::network::eip2718::Encodable2718;
 use alloy_rpc_types::{Block, BlockTransactions, Transaction};
 use pevm::{EvmAccount, PevmResult, PevmTxExecutionResult, Storage};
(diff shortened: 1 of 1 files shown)


tests/common/runner.rs
```diff
@@ -45,35 +45,35 @@ pub fn test_execute_revm<S: Storage + Clone + Send + Sync>(storage: S, txs: Vec<
     );
 }
 
+fn encode_receipt_2718(tx_type: u8, tx_result: &PevmTxExecutionResult) -> Bytes {
+    let tx_type = TxType::try_from(tx_type).unwrap();
+    let receipt_with_bloom = tx_result.receipt.clone().with_bloom();
```
hai-rise, line 50: Do we need anything else from `tx_result` for OP? If not I think passing in an owned `receipt` or `receipt_with_bloom` here would:
- Better Match the function name.
- Be tighter in case we need to encode receipts in a context without `PevmTxExecutionResult`.

tests/common/runner.rs
```diff
@@ -45,35 +45,35 @@ pub fn test_execute_revm<S: Storage + Clone + Send + Sync>(storage: S, txs: Vec<
...
-    // We use BTreeMap because the keys must be sorted in ascending order.
-    let trie_entries: BTreeMap<_, _> = receipt_envelope_iter
+        .map(|tx| tx.transaction_type.unwrap_or_default())
+        .zip(tx_results)
         .enumerate()
-        .map(|(index, receipt)| {
-            let key_buffer = alloy_rlp::encode_fixed_size(&index);
-            let mut value_buffer = Vec::new();
-            receipt.encode_2718(&mut value_buffer);
+        .map(|(index, (tx_type, tx_result))| {
```
hai-rise, line 74: Very nit, but `tx_idx` is more consistent with `tx_type`, `tx_result`, and the remaining code base!

tests/common/runner.rs
```diff
@@ -45,35 +45,35 @@ pub fn test_execute_revm<S: Storage + Clone + Send + Sync>(storage: S, txs: Vec<
     );
 }
 
+fn encode_receipt_2718(tx_type: u8, tx_result: &PevmTxExecutionResult) -> Bytes {
```
hai-rise, line 48: This can be done later when we add the `Chain` parameter, but this would be better put in `primitives` than here :thinking:?

tests/common/runner.rs
hai-rise: Looking back, we can even turn this into a `for`, and call `add_leaf` with the `key` & `value` calculated here without collecting them into an intermediate container first :thinking:.

tests/common/runner.rs
```diff
@@ -45,35 +45,35 @@ pub fn test_execute_revm<S: Storage + Clone + Send + Sync>(storage: S, txs: Vec<
...
-    let receipt_envelope_iter =
-        Iterator::zip(tx_type_iter, receipt_iter).map(|(tx_type, receipt)| match tx_type {
-            TxType::Legacy => ReceiptEnvelope::Legacy(receipt),
-            TxType::Eip2930 => ReceiptEnvelope::Eip2930(receipt),
-            TxType::Eip1559 => ReceiptEnvelope::Eip1559(receipt),
-            TxType::Eip4844 => ReceiptEnvelope::Eip4844(receipt),
-        });
-
-    // 2. Create a trie then calculate the root hash
-    // We use BTreeMap because the keys must be sorted in ascending order.
```
hai-rise, line 70: Let's not swallow this comment.

kien-rise:
Thanks to the appearane of https://github.com/alloy-rs/op-alloy, this PR is no longer needed.

For OP, we will use `OpReceiptEnvelope`.

PR #230, opened 2024-07-18
refactor: add trait PevmChain

kien-rise:

Later we will add `PevmChainOptimism`.


benches/gigagas.rs
```diff
@@ -7,10 +7,12 @@
 use std::{num::NonZeroUsize, thread};
 
 use ahash::AHashMap;
-use alloy_chains::Chain;
 use alloy_primitives::{Address, U160, U256};
 use criterion::{black_box, criterion_group, criterion_main, Criterion};
-use pevm::{execute_revm_parallel, execute_revm_sequential, EvmAccount, InMemoryStorage};
+use pevm::{
+    execute_revm_parallel, execute_revm_sequential, network::ethereum::PevmChainEthereum,
```
hai-rise, line 13: `network::PevmEthereum` would be cleaner than `network::ethereum::PevmChainEthereum`.

benches/gigagas.rs
```diff
@@ -30,7 +32,7 @@ static GLOBAL: rpmalloc::RpMalloc = rpmalloc::RpMalloc;
 
 pub fn bench(c: &mut Criterion, name: &str, state: common::ChainState, txs: Vec<TxEnv>) {
     let concurrency_level = thread::available_parallelism().unwrap_or(NonZeroUsize::MIN);
-    let chain = Chain::mainnet();
+    let chain = PevmChainEthereum::default();
```
hai-rise, line 35: `PevmEthereum::mainnet()` would be cleaner.

src/mv_memory.rs
```diff
@@ -9,31 +9,36 @@ use crate::{
...
+/// version of a corresponding transaction.
+#[derive(Debug)]
+pub struct MvMemory {
+    /// History of each location
     // No more hashing is required as we already identify memory locations by their hash
     // in the read & write sets. [dashmap] having a dedicated interface for this use case
     // (that skips hashing for [u64] keys) would make our code cleaner and "faster".
     // Nevertheless, the compiler should be good enough to optimize these cases anyway.
     data: DashMap<MemoryLocationHash, BTreeMap<TxIdx, MemoryEntry>, BuildIdentityHasher>,
+    /// Last locations
```
hai-rise, line 39: `Last read & written locations of each transaction`

src/mv_memory.rs
```diff
@@ -9,31 +9,36 @@ use crate::{
...
     last_locations: Vec<Mutex<LastLocations>>,
+    /// Lazy addresses
```
hai-rise, line 41: `Lazy addresses that need full evaluation at the end of the block`

src/network.rs
```diff
@@ -1,3 +1,39 @@
...
 
+use std::fmt::Debug;
+
+use alloy_primitives::U256;
+use alloy_rpc_types::{Header, Transaction};
+use revm::primitives::{BlockEnv, SpecId, TxEnv};
+
+use crate::mv_memory::MvMemory;
+
+/// A chain ID (u64) associated with relevant utils.
```
hai-rise, line 11: A trait doesn't define the full structure that implements it. A custom network may hold much more data than an id to, for example, build more efficient `MvMemory`. Something like `Custom behaviours for different chains & networks` would be better.

src/network.rs
```diff
@@ -1,3 +1,39 @@
...
+pub trait PevmChain {
```
hai-rise, line 12: This should be `PevmNetwork` as we are in `src/network`. If we go with `PevmChain` then this should be renamed to `src/chain`. I have no clear preference at the moment.

src/network.rs
```diff
@@ -1,3 +1,39 @@
...
+
+    /// Get chain id.
+    fn id(&self) -> u64;
+
+    /// Build [MvMemory]
+    fn build_mv_memory(
+        hasher: &ahash::RandomState,
+        block_env: &BlockEnv,
+        txs: &[TxEnv],
+    ) -> Result<MvMemory, Self::BuildMvMemoryError>;
```
hai-rise, line 30: Just force an `MvMemory` here.

src/network.rs
```diff
@@ -1,3 +1,39 @@
...
+
+    /// Get block's [SpecId]
+    fn get_block_spec(header: &Header) -> Result<SpecId, Self::GetBlockSpecError>;
```
hai-rise, line 33: `id` -> `get_block_spec` -> `get_gas_price` -> `build_mv_memory` is a better order.

src/network.rs
```diff
@@ -1,3 +1,39 @@
...
+    type GetBlockSpecError: Debug + Clone;
+
+    /// The error type for [Self::get_gas_price].
+    type GetGasPriceError: Debug + Clone;
+
+    /// Get chain id.
+    fn id(&self) -> u64;
+
+    /// Build [MvMemory]
+    fn build_mv_memory(
```
hai-rise, line 26: Let's provide a default implementation that returns an empty `MvMemory`. Trait implementors can then focus on chain specs to make an integration first before getting to the optimization stage. Facing concepts like `MvMemory`, lazy addresses, etc. too early would be bad DX. I think OP can just use this default implementation as well.

src/network/ethereum.rs
```diff
@@ -1,87 +1,143 @@
...
-    let mut lazy_addresses = LazyAddresses::default();
-    lazy_addresses.0.insert(block_env.coinbase);
-
-    MvMemory::new(block_size, estimated_locations, lazy_addresses)
+use super::PevmChain;
+
+/// Implementation of [PevmChain] for Ethereum
+#[derive(Debug, Clone)]
+pub struct PevmChainEthereum {
+    id: u64,
```
hai-rise, line 20: Not yet important but this not being `pub` without a `new` function would make it impossible to custom.

src/pevm.rs
```diff
@@ -34,7 +33,9 @@ use crate::{
 #[derive(Debug, PartialEq, Clone)]
 pub enum PevmError {
     /// Cannot derive the chain spec from the block header.
-    UnknownBlockSpec,
+    GetBlockSpecError(String),
```
hai-rise, line 36: We can parameterize `PevmError` by `PevmChain` to reuse its error types?

src/network.rs
hai-rise: All functions should have `&self` in case the implementing `struct` stores more information to use in the function implementations. For instance, I'd imagine different Ethereum networks having different hardfork timestamps for the same `SpecId`. A proper implementation would need to match on the chain id, or hardfork timestamps provided by the struct.

src/network/ethereum.rs
```diff
@@ -1,87 +1,143 @@
...
+            lazy_addresses,
+        ))
+    }
+
+    /// Get the REVM spec id of an Alloy block.
+    // Currently hardcoding Ethereum hardforks from these reference:
+    // https://github.com/paradigmxyz/reth/blob/4fa627736681289ba899b38f1c7a97d9fcf33dc6/crates/primitives/src/revm/config.rs#L33-L78
+    // https://github.com/paradigmxyz/reth/blob/4fa627736681289ba899b38f1c7a97d9fcf33dc6/crates/primitives/src/chain/spec.rs#L44-L68
+    // TODO: Better error handling & properly test this.
+    fn get_block_spec(header: &Header) -> Result<SpecId, Self::GetBlockSpecError> {
```
hai-rise, line 91: Should add a comment to `PevmEthereum` that we only support mainnet at the moment.

src/primitives.rs
```diff
@@ -29,7 +29,7 @@ pub(crate) fn get_block_env(header: &Header) -> Option<BlockEnv> {
 #[derive(Debug, Clone, PartialEq)]
 pub enum TransactionParsingError {
     OverflowedGasLimit,
-    MissingGasPrice,
+    GetGasPriceError(String),
```
hai-rise, line 32: We can parameterize `TransactionParsingError ` by `PevmChain` to reuse its error type?

src/vm.rs
```diff
@@ -102,8 +101,8 @@ pub(crate) enum VmExecutionResult {
 // structure & storage, and tracks the read set of the current execution.
 // TODO: Simplify this type, like grouping [from] and [to] into a
 // [preprocessed_addresses] or a [preprocessed_locations] vector.
-struct VmDb<'a, S: Storage> {
-    vm: &'a Vm<'a, S>,
+struct VmDb<'a, S: Storage, C: PevmChain> {
```
hai-rise, line 104: Looking back, propagating a lot of `PevmChain` here seems overkill. We can just pass in a `chain_id` instead?

src/vm.rs
```diff
@@ -489,26 +488,26 @@ impl<'a, S: Storage> Database for VmDb<'a, S> {
...
 
-impl<'a, S: Storage> Vm<'a, S> {
+impl<'a, S: Storage, C: PevmChain> Vm<'a, S, C> {
     pub(crate) fn new(
         hasher: &'a ahash::RandomState,
         storage: &'a S,
         mv_memory: &'a MvMemory,
         txs: &'a [TxEnv],
-        chain: Chain,
+        chain: &'a C,
```
hai-rise, line 510: Looking back, propagating a lot of PevmChain here seems overkill. We can just pass in a chain_id instead?

src/vm.rs
```diff
@@ -748,9 +747,9 @@ impl<'a, S: Storage> Vm<'a, S> {
     }
 }
 
-pub(crate) fn build_evm<'a, DB: Database>(
+pub(crate) fn build_evm<'a, DB: Database, C: PevmChain>(
     db: DB,
-    chain: Chain,
+    chain: &C,
```
hai-rise, line 752: Looking back, propagating a lot of PevmChain here seems overkill. We can just pass in a chain_id instead?

src/network/ethereum.rs
kien-rise: Adding `PevmChainEthereum::new()` will require to change `get_block_spec`.

I will add a comment to state that currently, only Ethereum Mainnet is supported.

src/vm.rs
kien-rise: as discussed, we need `PevmChain` in `build_evm`

src/chain.rs
```diff
@@ -0,0 +1,56 @@
...
+    fn get_gas_price(&self, tx: &Transaction) -> Result<U256, Self::GetGasPriceError>;
+
+    /// Build [MvMemory]
+    fn build_mv_memory(
+        &self,
+        hasher: &ahash::RandomState,
+        block_env: &BlockEnv,
+        txs: &[TxEnv],
+    ) -> MvMemory {
+        let block_size = txs.len();
```
hai-rise, line 38: This should be the Ethereum implementation. The default implementation should just return `MvMemory::new(txs.len(), HashMap::with_hasher(BuildIdentityHasher::default()), LazyAddresses::default())`.

src/chain/ethereum.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// [tx.type] is invalid.
+    InvalidType(u8),
+    /// [tx.gas_price] is none.
+    MissingGasPrice,
+    /// [tx.max_fee_per_gas] is none.
+    MissingMaxFeePerGas,
+}
+
+impl PevmChain for PevmEthereum {
+    type GetBlockSpecError = GetBlockSpecError;
```
hai-rise, line 49: Looking back, `BlockSpecError = EthereumBlockSpecError` is clearer.

src/chain/ethereum.rs
```diff
@@ -0,0 +1,114 @@
...
+    type GetGasPriceError = GetGasPriceError;
```
hai-rise, line 50: Looking back, `BlockGasPriceError = EthereumGasPriceError` is clearer.

src/mv_memory.rs
```diff
@@ -9,31 +9,36 @@ use crate::{
...
-// multiple writes for each memory location, along with a value and an associated
-// version of a corresponding transaction.
-pub(crate) struct MvMemory {
+/// The MvMemory contains shared memory in a form of a multi-version data
+/// structure for values written and read by different transactions. It stores
+/// multiple writes for each memory location, along with a value and an associated
+/// version of a corresponding transaction.
+#[derive(Debug)]
+pub struct MvMemory {
+    /// History of each location
```
hai-rise, line 33: `The list of transaction incarnations and written values for each memory location`

src/chain/ethereum.rs
kien-rise: Gas price is a property of a tx, not a block.

How about `TxGasPriceError`?

src/chain.rs
hai-rise: Which would be cleaner defined as a `Default` implementation in `mv_memory`, for us to just spam `MvMemory:Default()` here.

src/chain/ethereum.rs
hai-rise: Ah yes, just `GasPriceError` is good. Was a copy-paste mistake from `BlockSpecError`.

src/chain.rs
```diff
@@ -0,0 +1,56 @@
...
+use alloy_rpc_types::{Header, Transaction};
+use revm::primitives::{BlockEnv, SpecId, TxEnv};
+
+use crate::{
+    mv_memory::{LazyAddresses, MvMemory},
+    BuildIdentityHasher, MemoryLocation, TxIdx,
+};
+
+/// Custom behaviours for different chains & networks
+pub trait PevmChain: Debug + Clone + PartialEq {
```
hai-rise, line 15: Do we really need these `Debug + Clone + PartialEq`? If not then let's not restrict it. We never know if the actual struct implementing the trait would be easy/hard to compare or clone.

kien-rise: It seems that `Clone` is redundant but currently we need PartialEq so that we can compare two `PevmResult` values:

```
pub fn assert_execution_result<C: PevmChain>(
    sequential_result: &PevmResult<C>,
    parallel_result: &PevmResult<C>,
) {
    assert_eq!(sequential_result, parallel_result);
}
```

Should we compare only the `Ok` arms instead, and ignore if both `sequential_result` and and `parallel_result` are `Err`?

src/vm.rs
```diff
@@ -489,26 +488,26 @@ impl<'a, S: Storage> Database for VmDb<'a, S> {
...
 }
 
-pub(crate) struct Vm<'a, S: Storage> {
+pub(crate) struct Vm<'a, S: Storage, C: PevmChain> {
     hasher: &'a ahash::RandomState,
     storage: &'a S,
     mv_memory: &'a MvMemory,
     txs: &'a [TxEnv],
-    chain: Chain,
+    chain: &'a C,
```
hai-rise, line 496: Can you rebase and reorder this as `chain` -> `block_env` -> `txs`?

src/chain.rs
hai-rise: Wait we still need to pass `block_env` in. We can make two constructors or just write the long line here.

src/chain.rs
hai-rise: The most natural way is to restrict at `assert_execution_result` with something like:
```
pub fn assert_execution_result<C: PevmChain + PartialEq>(...
```

Meaning we can only assert execution results of chain objects that can be compared, but not all chain objects must be able to be compared.

Ideally, we wouldn't even need to require it as only the error types are represented in `PevmResults`, and it's fair to enforce these traits on error types.

kien-rise: The change is bigger than expected because we must implement `impl<C> PartialEq for PevmError<C>` manually. `#[derive(PartialEq)]` does not work. Please help to review commit https://github.com/risechain/pevm/pull/230/commits/a00024de6d1a72a9c9dc756e722bc3dad54be59e

src/chain.rs
kien-rise: Note that we cannot `impl Default for MvMemory` because `txs.len()` is required to initialize `last_locations`.

src/chain.rs
kien-rise: I updated the code. I only removed the trait bounds for `PevmChain`. You can compare the previous solution (https://github.com/risechain/pevm/commit/a00024de6d1a72a9c9dc756e722bc3dad54be59e) and the current solution (https://github.com/risechain/pevm/pull/230/commits/84aca8672c6cd81ed5eaada851c4d13566a725ba).

tests/mainnet.rs
```diff
@@ -49,11 +52,12 @@ fn mainnet_blocks_from_rpc() {
             )
             .unwrap()
             .unwrap();
-        let spec_id = ethereum::get_block_spec(&block.header).unwrap();
+        let chain = PevmEthereum::mainnet();
+        let spec_id = PevmEthereum::get_block_spec(&chain, &block.header).unwrap();
```
hai-rise, line 56: `chain.get_block_spec(&block.header)` is clearer?

Issue #75, opened 2024-05-15
Improve CI


hai-rise:
- Use GitHub action cache and more tricks to improve build speed.
- Add a feature flag to not run heavy (mainnet) tests in CI.

PR #233, opened 2024-07-24
feat: support optimism

kien-rise:
Resolves https://github.com/risechain/pevm/issues/60

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,14 @@ name = "pevm"
 version = "0.1.0"
 edition = "2021"
 
+[features]
+optimism = [
+    "revm/optimism",
+    "dep:op-alloy-consensus",
+    "dep:op-alloy-network",
+    "dep:op-alloy-rpc-types",
+]
+
 [dependencies]
 # Put this behind a feature flag if there are use cases & users
 # that need more security guarantees even for internally hashing
@@ -34,6 +42,11 @@ alloy-transport = "0.3.1"
 alloy-transport-http = "0.3.1"
 reqwest = "0.12.7"
 tokio = { version = "1.40.0", features = ["rt-multi-thread"] }
+
+# OP dependencies
+op-alloy-consensus = { version = "0.2.7", optional = true }
+op-alloy-network = { version = "0.2.7", optional = true }
+op-alloy-rpc-types = { version = "0.2.7", optional = true }
 
 [dev-dependencies]
 bincode = "1.3.3"
--- a/src/chain.rs
+++ b/src/chain.rs
@@ -18,6 +18,14 @@ use crate::{mv_memory::MvMemory, PevmTxExecutionResult};
 pub enum RewardPolicy {
     /// Ethereum
     Ethereum,
+    /// Optimism
+    #[cfg(feature = "optimism")]
+    Optimism {
+        /// L1 Fee Receipient
+        l1_fee_recipient_location_hash: crate::MemoryLocationHash,
+        /// Base Fee Vault
+        base_fee_vault_location_hash: crate::MemoryLocationHash,
+    },
 }
 
 /// The error type of [PevmChain::calculate_receipt_root]
@@ -93,3 +101,8 @@ pub trait PevmChain: Debug {
 
 mod ethereum;
 pub use ethereum::PevmEthereum;
+
+#[cfg(feature = "optimism")]
+mod optimism;
+#[cfg(feature = "optimism")]
+pub use optimism::PevmOptimism;
--- a/src/chain/ethereum.rs
+++ b/src/chain/ethereum.rs
@@ -151,6 +151,8 @@ impl PevmChain for PevmEthereum {
             blob_hashes: tx.blob_versioned_hashes.unwrap_or_default(),
             max_fee_per_blob_gas: tx.max_fee_per_blob_gas.map(U256::from),
             authorization_list: None, // TODO: Support in the upcoming hardfork
+            #[cfg(feature = "optimism")]
+            optimism: revm::primitives::OptimismFields::default(),
         })
     }
 
--- a/src/chain/optimism.rs
+++ b/src/chain/optimism.rs
@@ -1,0 +1,310 @@
+//! Optimism
+use std::collections::{BTreeMap, HashMap};
+
+use alloy_chains::NamedChain;
+use alloy_consensus::{Signed, TxEip1559, TxEip2930, TxEip4844, TxLegacy};
+use alloy_primitives::{Bytes, B256, U256};
+use alloy_rpc_types::{BlockTransactions, Header};
+use op_alloy_consensus::{OpDepositReceipt, OpReceiptEnvelope, OpTxEnvelope, OpTxType, TxDeposit};
+use op_alloy_network::eip2718::Encodable2718;
+use revm::{
+    primitives::{BlockEnv, OptimismFields, SpecId, TxEnv},
+    Handler,
+};
+
+use crate::{mv_memory::MvMemory, BuildIdentityHasher, MemoryLocation, PevmTxExecutionResult};
+
+use super::{CalculateReceiptRootError, PevmChain, RewardPolicy};
+
+/// Implementation of [PevmChain] for Optimism
+#[derive(Debug, Clone, PartialEq)]
+pub struct PevmOptimism {
+    id: u64,
+}
+
+impl PevmOptimism {
+    /// Optimism Mainnet
+    pub fn mainnet() -> Self {
+        PevmOptimism {
+            id: NamedChain::Optimism.into(),
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum OptimismBlockSpecError {
+    MissingBlockNumber,
+    UnsupportedSpec,
+}
+
+/// Represents errors that can occur when parsing transactions
+#[derive(Debug, Clone, PartialEq)]
+pub enum OptimismTransactionParsingError {
+    ConversionError(String),
+    InvalidType(u8),
+    MissingGasPrice,
+    MissingMaxFeePerGas,
+    MissingSourceHash,
+    OverflowedGasLimit,
+    SerdeError(String),
+}
+
+fn get_optimism_gas_price(
+    tx: &op_alloy_rpc_types::Transaction,
+) -> Result<U256, OptimismTransactionParsingError> {
+    let tx_type_raw = tx.inner.transaction_type.unwrap_or_default();
+    let Ok(tx_type) = OpTxType::try_from(tx_type_raw) else {
+        return Err(OptimismTransactionParsingError::InvalidType(tx_type_raw));
+    };
+
+    match tx_type {
+        OpTxType::Legacy | OpTxType::Eip2930 => tx
+            .inner
+            .gas_price
+            .map(U256::from)
+            .ok_or(OptimismTransactionParsingError::MissingGasPrice),
+        OpTxType::Eip1559 | OpTxType::Eip4844 => tx
+            .inner
+            .max_fee_per_gas
+            .map(U256::from)
+            .ok_or(OptimismTransactionParsingError::MissingMaxFeePerGas),
+        OpTxType::Deposit => Ok(U256::ZERO),
+    }
+}
+
+/// Convert [Transaction] to [OptimismFields]
+/// https://github.com/paradigmxyz/reth/blob/fc4c037e60b623b81b296fe9242fa905ff36b89a/crates/primitives/src/transaction/compat.rs#L99
+pub(crate) fn get_optimism_fields(
+    tx: &op_alloy_rpc_types::Transaction,
+) -> Result<OptimismFields, OptimismTransactionParsingError> {
+    let tx_type = tx.inner.transaction_type.unwrap_or_default();
+    let op_tx_type = OpTxType::try_from(tx_type)
+        .map_err(|_err| OptimismTransactionParsingError::InvalidType(tx_type))?;
+
+    let inner = tx.inner.clone();
+    let tx_envelope = match op_tx_type {
+        OpTxType::Legacy => Signed::<TxLegacy>::try_from(inner).map(OpTxEnvelope::from),
+        OpTxType::Eip2930 => Signed::<TxEip2930>::try_from(inner).map(OpTxEnvelope::from),
+        OpTxType::Eip1559 => Signed::<TxEip1559>::try_from(inner).map(OpTxEnvelope::from),
+        OpTxType::Eip4844 => Signed::<TxEip4844>::try_from(inner).map(OpTxEnvelope::from),
+        OpTxType::Deposit => {
+            let tx_deposit = TxDeposit {
+                source_hash: tx
+                    .source_hash
+                    .ok_or(OptimismTransactionParsingError::MissingSourceHash)?,
+                from: tx.inner.from,
+                to: tx.inner.to.into(),
+                mint: tx.mint,
+                value: tx.inner.value,
+                gas_limit: tx.inner.gas,
+                is_system_transaction: tx.is_system_tx.unwrap_or_default(),
+                input: tx.inner.input.clone(),
+            };
+            Ok(OpTxEnvelope::from(tx_deposit))
+        }
+    }
+    .map_err(|err| OptimismTransactionParsingError::ConversionError(err.to_string()))?;
+
+    let mut envelope_buf = Vec::<u8>::new();
+    tx_envelope.encode_2718(&mut envelope_buf);
+
+    Ok(OptimismFields {
+        source_hash: tx.source_hash,
+        mint: tx.mint,
+        is_system_transaction: tx.is_system_tx,
+        enveloped_tx: Some(Bytes::from(envelope_buf)),
+    })
+}
+
+impl PevmChain for PevmOptimism {
+    type Transaction = op_alloy_rpc_types::Transaction;
+    type BlockSpecError = OptimismBlockSpecError;
+    type TransactionParsingError = OptimismTransactionParsingError;
+
+    fn id(&self) -> u64 {
+        self.id
+    }
+
+    // TODO: allow to construct deposit transactions
+    fn build_tx_from_alloy_tx(&self, tx: alloy_rpc_types::Transaction) -> Self::Transaction {
+        Self::Transaction {
+            inner: tx,
+            mint: None,
+            source_hash: None,
+            is_system_tx: None,
+            deposit_receipt_version: None,
+        }
+    }
+
+    fn get_block_spec(&self, header: &Header) -> Result<SpecId, Self::BlockSpecError> {
+        // TODO: The implementation below is only true for Optimism Mainnet.
+        // When supporting other networks (e.g. Optimism Sepolia), remember to adjust the code here.
+        if header.timestamp >= 1720627201 {
+            Ok(SpecId::FJORD)
+        } else if header.timestamp >= 1710374401 {
+            Ok(SpecId::ECOTONE)
+        } else if header.timestamp >= 1704992401 {
+            Ok(SpecId::CANYON)
+        } else if header.number >= 105235063 {
+            // On Optimism Mainnet, Bedrock and Regolith are activated at the same time.
+            // Therefore, this function never returns SpecId::BEDROCK.
+            // The statement above might not be true for other network, e.g. Optimism Goerli.
+            Ok(SpecId::REGOLITH)
+        } else {
+            // TODO: revm does not support pre-Bedrock blocks.
+            // https://docs.optimism.io/builders/node-operators/architecture#legacy-geth
+            Err(OptimismBlockSpecError::UnsupportedSpec)
+        }
+    }
+
+    fn build_mv_memory(
+        &self,
+        hasher: &ahash::RandomState,
+        block_env: &BlockEnv,
+        txs: &[TxEnv],
+    ) -> MvMemory {
+        let beneficiary_location_hash = hasher.hash_one(MemoryLocation::Basic(block_env.coinbase));
+        let l1_fee_recipient_location_hash = hasher.hash_one(revm::L1_FEE_RECIPIENT);
+        let base_fee_recipient_location_hash = hasher.hash_one(revm::BASE_FEE_RECIPIENT);
+
+        // TODO: Estimate more locations based on sender, to, etc.
+        let mut estimated_locations = HashMap::with_hasher(BuildIdentityHasher::default());
+        for (index, tx) in txs.iter().enumerate() {
+            if tx.optimism.source_hash.is_none() {
+                estimated_locations
+                    .entry(beneficiary_location_hash)
+                    .or_insert_with(|| Vec::with_capacity(txs.len()))
+                    .push(index);
+            } else {
+                // TODO: Benchmark to check whether adding these estimated
+                // locations helps or harms the performance.
+                estimated_locations
+                    .entry(l1_fee_recipient_location_hash)
+                    .or_insert_with(|| Vec::with_capacity(1))
+                    .push(index);
+                estimated_locations
+                    .entry(base_fee_recipient_location_hash)
+                    .or_insert_with(|| Vec::with_capacity(1))
+                    .push(index);
+            }
+        }
+
+        MvMemory::new(
+            txs.len(),
+            estimated_locations,
+            [
+                block_env.coinbase,
+                revm::L1_FEE_RECIPIENT,
+                revm::BASE_FEE_RECIPIENT,
+            ],
+        )
+    }
+
+    fn get_handler<'a, EXT, DB: revm::Database>(
+        &self,
+        spec_id: SpecId,
+        with_reward_beneficiary: bool,
+    ) -> Handler<'a, revm::Context<EXT, DB>, EXT, DB> {
+        Handler::optimism_with_spec(spec_id, with_reward_beneficiary)
+    }
+
+    fn get_reward_policy(&self, hasher: &ahash::RandomState) -> RewardPolicy {
+        RewardPolicy::Optimism {
+            l1_fee_recipient_location_hash: hasher
+                .hash_one(MemoryLocation::Basic(revm::optimism::L1_FEE_RECIPIENT)),
+            base_fee_vault_location_hash: hasher
+                .hash_one(MemoryLocation::Basic(revm::optimism::BASE_FEE_RECIPIENT)),
+        }
+    }
+
+    // Refer to section 4.3.2. Holistic Validity in the Ethereum Yellow Paper.
+    // https://github.com/ethereum/go-ethereum/blob/master/cmd/era/main.go#L289
+    // https://github.com/paradigmxyz/reth/blob/b4a1b733c93f7e262f1b774722670e08cdcb6276/crates/primitives/src/proofs.rs
+    fn calculate_receipt_root(
+        &self,
+        spec_id: SpecId,
+        txs: &BlockTransactions<Self::Transaction>,
+        tx_results: &[PevmTxExecutionResult],
+    ) -> Result<B256, CalculateReceiptRootError> {
+        // 1. Create a list of [ReceiptEnvelope]
+        let receipt_envelopes: Vec<OpReceiptEnvelope> =
+            Iterator::zip(txs.txns(), tx_results.iter())
+                .map(|(tx, tx_result)| {
+                    let receipt = tx_result.receipt.clone();
+                    let byte = tx.inner.transaction_type.unwrap_or_default();
+                    let tx_type = OpTxType::try_from(byte)
+                        .map_err(|_| CalculateReceiptRootError::InvalidTxType(byte))?;
+                    Ok(match tx_type {
+                        OpTxType::Legacy => OpReceiptEnvelope::Legacy(receipt.with_bloom()),
+                        OpTxType::Eip2930 => OpReceiptEnvelope::Eip2930(receipt.with_bloom()),
+                        OpTxType::Eip1559 => OpReceiptEnvelope::Eip1559(receipt.with_bloom()),
+                        OpTxType::Eip4844 => OpReceiptEnvelope::Eip4844(receipt.with_bloom()),
+                        OpTxType::Deposit => {
+                            let account_maybe = tx_result
+                                .state
+                                .get(&tx.inner.from)
+                                .expect("Sender not found");
+                            let account = account_maybe.as_ref().expect("Sender not changed");
+                            let receipt = OpDepositReceipt {
+                                inner: receipt,
+                                deposit_nonce: (spec_id >= SpecId::CANYON)
+                                    .then_some(account.nonce - 1),
+                                deposit_receipt_version: (spec_id >= SpecId::CANYON).then_some(1),
+                            };
+                            OpReceiptEnvelope::Deposit(receipt.with_bloom())
+                        }
+                    })
+                })
+                .collect::<Result<_, _>>()?;
+
+        // 2. Create a trie then calculate the root hash
+        // We use [BTreeMap] because the keys must be sorted in ascending order.
+        let trie_entries: BTreeMap<_, _> = receipt_envelopes
+            .iter()
+            .enumerate()
+            .map(|(index, receipt)| {
+                let key_buffer = alloy_rlp::encode_fixed_size(&index);
+                let mut value_buffer = Vec::new();
+                receipt.encode_2718(&mut value_buffer);
+                (key_buffer, value_buffer)
+            })
+            .collect();
+
+        let mut hash_builder = alloy_trie::HashBuilder::default();
+        for (k, v) in trie_entries {
+            hash_builder.add_leaf(alloy_trie::Nibbles::unpack(&k), &v);
+        }
+        Ok(hash_builder.root())
+    }
+
+    fn get_tx_env(&self, tx: Self::Transaction) -> Result<TxEnv, OptimismTransactionParsingError> {
+        Ok(TxEnv {
+            optimism: get_optimism_fields(&tx)?,
+            caller: tx.inner.from,
+            gas_limit: tx
+                .inner
+                .gas
+                .try_into()
+                .map_err(|_| OptimismTransactionParsingError::OverflowedGasLimit)?,
+            gas_price: get_optimism_gas_price(&tx)?,
+            gas_priority_fee: tx.inner.max_priority_fee_per_gas.map(U256::from),
+            transact_to: tx.inner.to.into(),
+            value: tx.inner.value,
+            data: tx.inner.input,
+            nonce: Some(tx.inner.nonce),
+            chain_id: tx.inner.chain_id,
+            access_list: tx.inner.access_list.unwrap_or_default().into(),
+            blob_hashes: tx.inner.blob_versioned_hashes.unwrap_or_default(),
+            max_fee_per_blob_gas: tx.inner.max_fee_per_blob_gas.map(U256::from),
+            authorization_list: None, // TODO: Support in the upcoming hardfork
+        })
+    }
+
+    fn is_eip_1559_enabled(&self, _spec_id: SpecId) -> bool {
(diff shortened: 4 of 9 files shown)


examples/check-optimism.rs
```diff
@@ -0,0 +1,119 @@
...
+    #[arg(long, env, default_value = "https://mainnet.optimism.io")]
+    rpc_url: String,
+
+    /// Block number
+    #[arg()]
+    block_number: u64,
+}
+
+#[cfg(feature = "optimism")]
+fn main() -> Result<(), Box<dyn std::error::Error>> {
```
hai-rise, line 33: Can we turn this into a test?

Cargo.toml
```diff
@@ -3,6 +3,10 @@ name = "pevm"
 version = "0.1.0"
 edition = "2021"
 
+[features]
+optimism = ["revm/optimism"]
+default = ["optimism"]
```
hai-rise, line 8: This should not be a default feature.

src/chain/optimism.rs
```diff
@@ -0,0 +1,266 @@
...
+        hasher: &ahash::RandomState,
+        block_env: &BlockEnv,
+        txs: &[TxEnv],
+    ) -> MvMemory {
+        let beneficiary_location_hash = hasher.hash_one(MemoryLocation::Basic(block_env.coinbase));
+
+        // TODO: Estimate more locations based on sender, to, etc.
+        let mut estimated_locations = HashMap::with_hasher(BuildIdentityHasher::default());
+        estimated_locations.insert(
+            beneficiary_location_hash,
```
hai-rise, line 207: What about `revm::L1_FEE_RECIPIENT` and `revm::BASE_FEE_RECIPIENT`?

src/primitives.rs
```diff
@@ -43,6 +48,10 @@ pub(crate) fn get_tx_env<C: PevmChain>(
     tx: Transaction,
 ) -> Result<TxEnv, TransactionParsingError<C>> {
     Ok(TxEnv {
+        #[cfg(feature = "optimism")]
```
hai-rise, line 51: Put this last?

src/storage/rpc.rs
```diff
@@ -70,6 +70,20 @@ impl<N> RpcStorage<N> {
     pub fn get_cache_block_hashes(&self) -> AHashMap<u64, B256> {
         self.cache_block_hashes.lock().unwrap().clone()
     }
+
+    /// Update the cache of accounts
```
hai-rise, line 74: Do we really need this? Can the call site just call the get function and apply their changes externally?

tests/ethereum/main.rs
```diff
@@ -61,6 +61,9 @@ fn build_block_env(env: &Env) -> BlockEnv {
 
 fn build_tx_env(tx: &TransactionParts, indexes: &TxPartIndices) -> Result<TxEnv, ParseError> {
     Ok(TxEnv {
+        #[cfg(feature = "optimism")]
```
hai-rise, line 64: Put it last?

src/chain.rs
```diff
@@ -18,24 +18,38 @@ use crate::{mv_memory::MvMemory, PevmTxExecutionResult};
 pub enum RewardPolicy {
     /// Ethereum
     Ethereum,
+    /// Optimism
+    #[cfg(feature = "optimism")]
+    Optimism {
+        /// L1 Fee Receipient
+        l1_fee_recipient_location_hash: crate::MemoryLocationHash,
```
hai-rise, line 25: We already have `use crate::{` above so let's just add `MemoryLocationHash` there.

src/chain.rs
```diff
@@ -18,24 +18,38 @@ use crate::{mv_memory::MvMemory, PevmTxExecutionResult};
...
+        /// Base Fee Vault
+        base_fee_vault_location_hash: crate::MemoryLocationHash,
+    },
 }
 
 /// Custom behaviours for different chains & networks
 pub trait PevmChain: Debug {
+    /// The transaction type
+    type Transaction: Debug + Clone + PartialEq;
```
hai-rise, line 34: Let's port this, `TxEnvError`, and `get_tx_env` into a dedicated PR to merge first.

src/chain/optimism.rs
```diff
@@ -0,0 +1,291 @@
...
+        for (index, tx) in txs.iter().enumerate() {
+            if tx.optimism.source_hash.is_none() {
+                estimated_locations
+                    .entry(beneficiary_location_hash)
+                    .or_insert_with(|| Vec::with_capacity(txs.len()))
+                    .push(index);
+            } else {
+                estimated_locations
+                    .entry(l1_fee_recipient_location_hash)
+                    .or_insert_with(|| Vec::with_capacity(1))
```
hai-rise, line 204: This can be `vec![index]`?

src/chain.rs
kien-rise: Note that this part is under `#[cfg(feature = "optimism")]`.

src/storage/rpc.rs
```diff
@@ -1,7 +1,7 @@
 // TODO: Put this behind an RPC flag to not pollute the core
 // library with RPC network & transport dependencies, etc.
 
-use std::{fmt::Debug, future::IntoFuture, sync::Mutex};
+use std::{fmt::Debug, future::IntoFuture, sync::Mutex, time::Duration};
```
hai-rise, line 4: Let's move these changes to a future PR. Can extend #75 to cover this.

tests/mainnet.rs
```diff
@@ -24,7 +24,7 @@ fn mainnet_blocks_from_rpc() {
     let rpc_url = match std::env::var("RPC_URL") {
         // The empty check is for GitHub Actions where the variable is set with an empty string when unset!?
         Ok(value) if !value.is_empty() => value.parse().unwrap(),
-        _ => Url::parse("https://eth.llamarpc.com").unwrap(),
+        _ => Url::parse("https://eth.public-rpc.com").unwrap(),
```
hai-rise, line 27: Let's merge this one line as a separate PR first if it is critical (recent community PRs did fail RPC).

src/chain.rs
kien-rise: if we add to `use crate::{`, then we will have to add something like `#[cfg_attr(not(feature = "optimism"), allow(unused_variables))]`

tests/optimism.rs
```diff
@@ -0,0 +1,125 @@
...
+        );
+        let parallel_result = pevm.execute(
+            &pre_state_rpc_storage,
+            &chain,
+            block.clone(),
+            concurrency_level,
+            false,
+        );
+
+        assert_eq!(&sequential_result, &parallel_result);
```
hai-rise, line 63: Can we share these checks with `common`/`mainnet`?

src/chain.rs
kien-rise: Will be resolved in https://github.com/risechain/pevm/pull/331

tests/mainnet.rs
kien-rise: To be resolved in https://github.com/risechain/pevm/pull/332

src/storage/rpc.rs
kien-rise: Let's see if the CI in #332 encounters any 429 errors.

kien-rise: To be resolved in https://github.com/risechain/pevm/pull/333

kien-rise:
GitHub keeps closing PRs when using "magic" words. :(

tests/optimism.rs
kien-rise: I haven't found any meaningful abstraction yet.

The most reasonable plan to me is to change to:

```
assert_eq!(
            block.header.logs_bloom,
            PevmTxExecutionResult::get_logs_bloom(&tx_results)
);

assert_eq!(
            block.header.gas_used,
            PevmTxExecutionResult::get_cumulative_gas_used(&tx_results)
);
```

(or replace `PevmTxExecutionResult::` by `common::` if we don't want to pollute our lib).

------

Another way is to implement `assert_header(&block.header, &tx_results)`. This fn compares `block.header.receipts_root`, `block.header.logs_bloom`, `block.header.gas_used` to the `tx_results`. However, this fn name be misleading because we are only checking a subset of header.

In this case, I prefer WET (write everything twice) to DRY (don't repeat yourself), until we can find a meaningful abstraction.

src/chain.rs
hai-rise: Ah that's more annoying, so yeah, let's keep this as is.

tests/optimism.rs
hai-rise: Being ad-hoc is fine for less important chains. Here OP (i.e., RISE as well) and Ethereum are of the same importance. It's weird to see one having on-disk checks while the others don't, then the other having `observed_accounts` and `expected_storage` checks that one doesn't, etc. The whole structure, flow, and what to check should be very similar. 

This is not trivial, so perhaps we can start with some mock blocks (like early day Ethereum), then complete OP tests in the upcoming PRs.

tests/optimism.rs
```diff
@@ -0,0 +1,125 @@
...
+        let tx_results = sequential_result.unwrap();
+
+        // We can only calculate the receipts root from Byzantium.
```
hai-rise, line 66: Is this even applicable to Optimism?

kien-rise: It is not applicable to Optimism because the min spec is `BEDROCK = 16` which is >= than `BYZANTIUM = 6`.

src/chain.rs
```diff
@@ -67,7 +75,15 @@ pub trait PevmChain: Debug {
         txs: &BlockTransactions<Self::Transaction>,
         tx_results: &[PevmTxExecutionResult],
     ) -> B256;
+
+    /// Build Self::Transaction type from Alloy's transaction
```
hai-rise, line 79: Let's move this and the test changes to a PR to merge first.
Also put this below `id`.

tests/common/mod.rs
```diff
@@ -4,15 +4,22 @@ use std::{
...
+use alloy_primitives::{uint, Address, Bloom, Bytes, B256, U256};
+use alloy_rpc_types::{Block, BlockTransactions, Header, Signature};
+use pevm::{chain::PevmChain, BlockHashes, Bytecodes, EvmAccount, InMemoryStorage};
 
 pub mod runner;
 pub use runner::{mock_account, test_execute_alloy, test_execute_revm};
 pub mod storage;
 
-pub static MOCK_ALLOY_BLOCK_HEADER: Header = Header {
+const MOCK_SIGNATURE: Signature = Signature {
```
hai-rise, line 15: Minor but let's put this below `MOCK_ALLOY_BLOCK_HEADER`.

tests/common/mod.rs
```diff
@@ -80,3 +87,42 @@ pub fn for_each_block_from_disk(mut handler: impl FnMut(Block, InMemoryStorage))
         );
     }
 }
+
+pub fn test_independent_raw_transfers<C>(chain: &C, block_size: usize)
```
hai-rise, line 91:
```
// Test a chain with [block_size] independent raw transactions that transfer to itself
```

tests/common/mod.rs
```diff
@@ -80,3 +87,42 @@ pub fn for_each_block_from_disk(mut handler: impl FnMut(Block, InMemoryStorage))
...
+where
+    C: PevmChain + Send + Sync + PartialEq,
+    C::Transaction: Default,
+{
+    let accounts: Vec<(Address, EvmAccount)> = (0..block_size).map(mock_account).collect();
+
+    // Create a block containing N raw-transfer transactions.
```
hai-rise, line 98: This can be removed after adding the function comment.

tests/common/mod.rs
```diff
@@ -80,3 +87,42 @@ pub fn for_each_block_from_disk(mut handler: impl FnMut(Block, InMemoryStorage))
...
+                        signature: Some(MOCK_SIGNATURE),
+                        ..alloy_rpc_types::Transaction::default()
+                    })
+                })
+                .collect(),
+        ),
+        ..Block::<C::Transaction>::default()
+    };
+
+    // Create a [InMemoryStorage] including the senders and the mock beneficiary account (`Address::ZERO`).
```
hai-rise, line 124: Can remove this now.

tests/raw_transfers.rs
```diff
@@ -68,33 +67,17 @@ fn raw_transfers_same_sender_multiple_txs() {
...
-                        }
-                    })
-                    .collect(),
-            ),
-            ..Block::default()
-        },
-        false,
-    );
+fn ethereum_independent_raw_transfers() {
+    common::test_independent_raw_transfers(&PevmEthereum::mainnet(), 0);
```
hai-rise, line 71: For better context, we should either:
- Keep these two edge cases in their dedicated tests.
- Add comments here on why we test for 0 and 1 transactiron.

src/chain.rs
```diff
@@ -67,7 +75,15 @@ pub trait PevmChain: Debug {
...
+    fn cast_from_eth_transaction(&self, tx: alloy_rpc_types::Transaction) -> Self::Transaction;
```
hai-rise, line 80: `eth_transaction` is wrong. For instance, Ethereum transactions don't have the `signature` that `alloy_rpc_types::Transaction` supports. The input here can keep growing to cover more chains, and that's why it can be converted to an arbitrary `Self::Transaction` here.

The more precise name would be `build_tx_from_{alloy|rpc}_tx`.

src/chain.rs
kien-rise: See https://github.com/risechain/pevm/pull/344

tests/raw_transfers.rs
kien-rise: > Add comments here on why we test for 0 and 1 transactiron

Edge cases are always worth testing. I think we don't need any comments to explain the "why". The function names are already self-explanatory.

src/chain.rs
```diff
@@ -74,3 +82,8 @@ pub trait PevmChain: Debug {
 
 mod ethereum;
 pub use ethereum::PevmEthereum;
+
+#[cfg(feature = "optimism")]
+pub(crate) mod optimism;
```
hai-rise, line 87: Can we remove `pub(crate)` like `ethereum` above?

src/chain/ethereum.rs
```diff
@@ -151,6 +151,8 @@ impl PevmChain for PevmEthereum {
             blob_hashes: tx.blob_versioned_hashes.unwrap_or_default(),
             max_fee_per_blob_gas: tx.max_fee_per_blob_gas.map(U256::from),
             authorization_list: None, // TODO: Support in the upcoming hardfork
+            #[cfg(feature = "optimism")]
+            optimism: revm::primitives::OptimismFields::default(),
```
hai-rise, line 155: This is still very annoying, but we must follow `revm` for now sigh.

src/chain/optimism.rs
```diff
@@ -0,0 +1,286 @@
+//! Optimism
+#![allow(missing_docs)]
```
hai-rise, line 2: Why :smiling_face_with_tear:?

src/chain/optimism.rs
```diff
@@ -0,0 +1,286 @@
...
+#[derive(Debug, Clone, PartialEq)]
+pub enum OptimismTransactionParsingError {
+    ConversionError(String),
+    InvalidType(u8),
+    MissingGasPrice,
+    MissingMaxFeePerGas,
+    MissingSourceHash,
+    OverflowedGasLimit,
+    SerdeError(String),
+    UnexpectedType(u8),
```
hai-rise, line 51: Can we merge this with `InvalidType`?

src/chain/optimism.rs
```diff
@@ -0,0 +1,286 @@
...
+            .inner
+            .max_fee_per_gas
+            .map(U256::from)
+            .ok_or(OptimismTransactionParsingError::MissingMaxFeePerGas),
+        OpTxType::Deposit => Ok(U256::ZERO),
+    }
+}
+
+/// Convert [Transaction] to [OptimismFields]
+pub(crate) fn get_optimism_fields(
```
hai-rise, line 76: These functions are non-trivial. Let's add a URL to OP's doc or an implementation in op-geth/op-reth/alloy/etc as reference.

src/chain/optimism.rs
```diff
@@ -0,0 +1,286 @@
...
+    Ok(OptimismFields {
+        source_hash: tx.source_hash,
+        mint: tx.mint,
+        is_system_transaction: tx.is_system_tx,
+        enveloped_tx: Some(envelope_buf),
+    })
+}
+
+impl PevmChain for PevmOptimism {
+    type Transaction = Transaction;
```
hai-rise, line 121: Let's qualify this.

src/chain/optimism.rs
```diff
@@ -0,0 +1,286 @@
...
+    type TransactionParsingError = OptimismTransactionParsingError;
+
+    fn id(&self) -> u64 {
+        self.id
+    }
+
+    fn build_tx_from_alloy_tx(&self, tx: alloy_rpc_types::Transaction) -> Self::Transaction {
+        Transaction {
+            inner: tx,
+            ..Transaction::default()
```
hai-rise, line 132: While this is okay for testing, I wonder if this works for real usage. Maybe it's safer to list the remaining fields with a comment why these can be default.

src/chain/optimism.rs
```diff
@@ -0,0 +1,286 @@
...
+        }
+    }
+
+    fn get_block_spec(&self, header: &Header) -> Result<SpecId, Self::BlockSpecError> {
```
hai-rise, line 136: Let's add a comment to support RISE (we'll do it soon) and other OP testnets (community contributions).

src/chain/optimism.rs
kien-rise: The remaining fields are `Option<_>`, which are always `None` unless the transaction is a Deposit Transaction (tx type = 126).

So, the fn implementation is realistic.

src/chain/optimism.rs
```diff
@@ -0,0 +1,286 @@
...
+        if header.timestamp >= 1720627201 {
+            Ok(SpecId::FJORD)
+        } else if header.timestamp >= 1710374401 {
+            Ok(SpecId::ECOTONE)
+        } else if header.timestamp >= 1704992401 {
+            Ok(SpecId::CANYON)
+        } else if header.number >= 105235063 {
+            Ok(SpecId::REGOLITH)
+        } else {
+            // TODO: revm does not support when L1Block is not available
```
hai-rise, line 146: There's no default value like `BEDROCK`? Maybe a link to an `revm` issue or code would help clarify this.

src/chain/optimism.rs
```diff
@@ -0,0 +1,286 @@
...
+        hasher: &ahash::RandomState,
+        block_env: &BlockEnv,
+        txs: &[TxEnv],
+    ) -> MvMemory {
+        let beneficiary_location_hash = hasher.hash_one(MemoryLocation::Basic(block_env.coinbase));
+        let l1_fee_recipient_location_hash = hasher.hash_one(revm::L1_FEE_RECIPIENT);
+        let base_fee_recipient_location_hash = hasher.hash_one(revm::BASE_FEE_RECIPIENT);
+
+        // TODO: Estimate more locations based on sender, to, etc.
+        let mut estimated_locations = HashMap::with_hasher(BuildIdentityHasher::default());
```
hai-rise, line 162: Let's add a `TODO` to benchmark without pre-estimating L1 fee & base fee recipient locations. If they are small pre-processing in sequential mode can be slower than dynamically inserting them in parallel threads.

src/chain/optimism.rs
```diff
@@ -0,0 +1,286 @@
...
+    // Refer to section 4.3.2. Holistic Validity in the Ethereum Yellow Paper.
+    // https://github.com/ethereum/go-ethereum/blob/master/cmd/era/main.go#L289
+    // https://github.com/paradigmxyz/reth/blob/b4a1b733c93f7e262f1b774722670e08cdcb6276/crates/primitives/src/proofs.rs
+    fn calculate_receipt_root(
+        &self,
+        spec_id: SpecId,
+        txs: &BlockTransactions<Self::Transaction>,
+        tx_results: &[PevmTxExecutionResult],
+    ) -> B256 {
+        // 1. Create an iterator of ReceiptEnvelope
```
hai-rise, line 218: Very minor `[OpReceiptEnvelope]`.

src/vm.rs
```diff
@@ -683,18 +684,52 @@ impl<'a, S: Storage, C: PevmChain> Vm<'a, S, C> {
     }
 
     // Apply rewards (balance increments) to beneficiary accounts, etc.
-    fn apply_rewards(&self, write_set: &mut WriteSet, tx: &TxEnv, gas_used: U256) {
+    fn apply_rewards<DB: Database>(
+        &self,
+        write_set: &mut WriteSet,
+        tx: &TxEnv,
+        gas_used: U256,
+        #[allow(unused_variables)] evm_context: &EvmContext<DB>,
```
hai-rise, line 692: Or we can put this under the OP feature flag (for now)? Also to be safe in case the Rust compiler doesn't remove the extra unused argument in Ethereum mode.

src/vm.rs
```diff
@@ -683,18 +684,52 @@ impl<'a, S: Storage, C: PevmChain> Vm<'a, S, C> {
...
+    ) {
+        let mut gas_price = if let Some(priority_fee) = tx.gas_priority_fee {
+            std::cmp::min(tx.gas_price, priority_fee + self.block_env.basefee)
+        } else {
+            tx.gas_price
+        };
+        if self.spec_id.is_enabled_in(SpecId::LONDON) {
```
hai-rise, line 699: We should add feature flags to all these spec checks? Not only for performance but to also avoid funny spec clashes between chains.

src/vm.rs
```diff
@@ -683,18 +684,52 @@ impl<'a, S: Storage, C: PevmChain> Vm<'a, S, C> {
...
+            gas_price = gas_price.saturating_sub(self.block_env.basefee);
+        }
+
         let rewards: SmallVec<[(MemoryLocationHash, U256); 1]> = match self.reward_policy {
```
hai-rise, line 703: We can make it `3` for OP here. Or even remove `SmallVec` and use array of 1 or 3 based on feature flag.

src/vm.rs
```diff
@@ -683,18 +684,52 @@ impl<'a, S: Storage, C: PevmChain> Vm<'a, S, C> {
...
+            } => {
+                let is_deposit = tx.optimism.source_hash.is_some();
+                if is_deposit {
+                    SmallVec::new()
                 } else {
-                    tx.gas_price
-                };
-                if self.spec_id.is_enabled_in(SpecId::LONDON) {
-                    gas_price = gas_price.saturating_sub(self.block_env.basefee);
+                    let Some(enveloped_tx) = &tx.optimism.enveloped_tx else {
```
hai-rise, line 716: Let's add a `// TODO: Better error handling`.

src/vm.rs
```diff
@@ -683,18 +684,52 @@ impl<'a, S: Storage, C: PevmChain> Vm<'a, S, C> {
...
+        }
+
         let rewards: SmallVec<[(MemoryLocationHash, U256); 1]> = match self.reward_policy {
             RewardPolicy::Ethereum => {
-                let mut gas_price = if let Some(priority_fee) = tx.gas_priority_fee {
-                    std::cmp::min(tx.gas_price, priority_fee + self.block_env.basefee)
+                smallvec![(self.beneficiary_location_hash, gas_price * gas_used)]
+            }
+            #[cfg(feature = "optimism")]
+            RewardPolicy::Optimism {
```
hai-rise, line 708: A link to these fee formulas would help.

src/chain/optimism.rs
hai-rise: How is it realistic when you just name a case it handles wrong.. We should at least add a `TODO` to address that, or note why we don't support it.

kien-rise: `alloy_rpc_types::Transaction` can never express a Deposit Transaction, because it does not have the dedicated OP fields (`mint`, `source_hash`, `is_system_tx`, `deposit_receipt_version`).

In other words, there is no way a user can (accidentally or intentionally) create a Deposit Transaction of type `alloy_rpc_types::Transaction`.

Therefore, a `TODO:` is not needed.

-----

We can expand the `..Transaction::default()`, just to make it more verbose.

kien-rise: The code will look like this:

```
fn build_tx_from_alloy_tx(&self, tx: alloy_rpc_types::Transaction) -> Self::Transaction {
        Self::Transaction {
            inner: tx,
            mint: None,
            source_hash: None,
            is_system_tx: None,
            deposit_receipt_version: None,
        }
    }

```

hai-rise: Okay, let's keep the current code and add a `TODO` to support building a deposit transaction. For instance, by: 
- Add a `TxExtraFields` type to `PevmChain` with a `Default` constraint. It is `()` for Ethereum and a struct of four fields for OP.
- Convert this function to `fn build_tx(&self, base: alloy_rpc_types::Transaction, extra: Self::TxExtraFields) Self::Transaction`.

It is of low priority until we write tests with deposits, even then we can construct them manually.

src/vm.rs
kien-rise: To me it feels like the same mistake when the `optimism` field is added to `struct TxEnv`:

```rust
    // Apply rewards (balance increments) to beneficiary accounts, etc.
    fn apply_rewards<DB: Database>(
        &self,
        write_set: &mut WriteSet,
        tx: &TxEnv,
        gas_used: U256,
        #[cfg(feature = "optimism")] evm_context: &EvmContext<DB>,
    )

----

              self.apply_rewards(
                &mut write_set,
                tx,
                U256::from(result_and_state.result.gas_used()),
                #[cfg(feature = "optimism")]
                &evm.context.evm,
            );
```


Compare the above to the cursed `optimism` field below:

```rust
    fn get_tx_env(&self, tx: Self::Transaction) -> Result<TxEnv, EthereumTransactionParsingError> {
        Ok(TxEnv {
            caller: tx.from,
            gas_limit: tx
                .gas
                .try_into()
                .map_err(|_| EthereumTransactionParsingError::OverflowedGasLimit)?,
            gas_price: get_ethereum_gas_price(&tx)?,
            gas_priority_fee: tx.max_priority_fee_per_gas.map(U256::from),
            transact_to: tx.to.into(),
            value: tx.value,
            data: tx.input,
            nonce: Some(tx.nonce),
            chain_id: tx.chain_id,
            access_list: tx.access_list.unwrap_or_default().into(),
            blob_hashes: tx.blob_versioned_hashes.unwrap_or_default(),
            max_fee_per_blob_gas: tx.max_fee_per_blob_gas.map(U256::from),
            authorization_list: None, // TODO: Support in the upcoming hardfork
            #[cfg(feature = "optimism")]
            optimism: revm::primitives::OptimismFields::default(),
        })
    }
```



-----

Personally, I think a feature flag should only:
1. Add a definition (function, module)
2. Add an arm to an enum

A feature flag should not:
1. Change an interface (including a function signature)

(The feature flags related to OS, CPU can be considered as exceptions. They can conditionally dictate the implementations.)

src/vm.rs
kien-rise: That's something we can consider in future PRs.

I don't expect that removing an `if` statement improves the performance significantly.

If any benchmark proves the performance improvement, then yes we should do it. Otherwise it will just add noise to `Cargo.toml` and other `*.rs` files.

src/vm.rs
hai-rise: How is adding an arm to an enum different than adding a parameter to a function? You have conditional compilation both at the definition and call sites (matches for `enum`).

If we don't want the extra arms unnecessarily increase all enum variants' memory footprint, then we also don't want the extra function parameters to increase functions' memory footprint when we're sure we don't use them.

src/vm.rs
kien-rise: I think we shouldn't.

In one block, there will only be _a single tx_ that uses 3 elements (to be more specific, the deposit transaction). The remaining N-1 blocks only uses _1_ element here.

src/chain/optimism.rs
kien-rise: I added. Feel free to suggest a change if you want.

src/vm.rs
hai-rise: No. Technically, there's no `LONDON` nor `SPURIOUS_DRAGON` in the context of OP and RISE. We'll even use a different Merkle tree for chain state with different state expiry rules, etc. 

Ignoring it will lead to bugs like skipping receipts root check for the first 4.3 millions blocks on new chains:
https://github.com/risechain/pevm/blob/3449a629ab3480ea2b4c57e710101f5dab751552/tests/common/runner.rs#L65-L74

These `>` checks are extremely slippery and not necessarily easy to notice when one mis-hits. An L2 can still proceed when the sequencer proposes a wrong state so we'll keep going when we mis-calculate a fee, receipt root, mis-clear a slot, etc., only for someone to challenge the wrong state transition and revert the chain plus fines.

So, `revm` still compiling here is a hack if anything. We (and users) must have full control of which chain applies which logic, including moving more spec abstractions towards our `Chain`. We're supposed to depend on Alloy types only and can support different underlying executors anyway.

A quick fix now is to use the `optimism` feature flag to either skip the check, or to use the correct spec (e.g., `BEDROCK` for OP.

src/vm.rs
kien-rise: added a link to @bluealloy/revm

src/vm.rs
hai-rise: You just contradicted yourself. If a `match` didn't have a default branch now they need to add one, just like how one need to add `..Default::default()` to the struct constructions. 

The main difference with `TxEnv` is that it is a public (and commonly used) type, so the pollution is annoying. `apply_rewards` is private and only called internally once, which is even less pervasive than the `pub` `RewardPolicy`.

src/vm.rs
kien-rise: Not relevant to this PR, but this line:

```
if block.header.number >= 4370000
```

should be changed to:

```
if spec_id >= SpecId::BYZANTIUM
```

src/vm.rs
hai-rise: Okay, can keep this as is then. Can try `[MemoryLocationHash; 1]` for Ethereum later.

src/vm.rs
kien-rise: Let's discuss offline tomorrow. I am not convinced.

> Technically, there's no LONDON nor SPURIOUS_DRAGON in the context of OP and RISE.

means `if self.spec_id.is_enabled_in(SpecId::LONDON) {` is an if statement where the condition is always true (because `SpecId::BEDROCK`, `SpecId::REGOLITH` are greater than `SpecId::LONDON`). So everything is fine!


> We (and users) must have full control of which chain applies which logic, including moving more spec abstractions towards our Chain.

Yes, if a chain has a specific logic, the logic should be put in `Chain`, not in the feature flag. Note that it's totally fine if someone compile with `--all-features` but then use `PevmEthereum::mainnet()`.

hai-rise: Which we still need to make sure is only enabled for Ethereum. `BYZANTIUM` is a 6 so a new chain somehow starting from `0` and misputs it here is a risk we don't want.

src/chain/optimism.rs
kien-rise: `SpecId::BEDROCK` exists but `revm` is not handling a Bedrock block correctly. Therefore, I throw an error here to stop execution immediately.

Let me verify again the `revm` issue when handling a Bedrock block.

src/vm.rs
kien-rise: > `apply_rewards` is private

ok, this point makes sense

src/vm.rs
hai-rise: > means if self.spec_id.is_enabled_in(SpecId::LONDON) { is an if statement where the condition is always true (because SpecId::BEDROCK, SpecId::REGOLITH are greater than SpecId::LONDON). So everything is fine!

This literally means everything is not fine lol. I repeat this is a bad hack from `revm` because there's no linearity between chains' hardforks. What if RISE picks `SpecId::LATEST` but doesn't inherit the state removal of `SPURIOUS_DRAGON`? What if we launch with (some) `ECOTONE` specs of OP and some `PRAGUE` of Ethereum? Not to mention the risk of `revm` reordering things around and things still compile. 

Checking for `LONDON` when we have a feature flag to pick `BEDROCK` and saying "everything is fine because the condition is always true" is insane.

src/chain/optimism.rs
hai-rise: That sounds like something we should definitely have some doc for.

src/vm.rs
hai-rise: Wait, the actual fix is to add to `Chain` things like `has_eip_161`, `has_eip_658`, etc. that take in a `SpecId`. The problem (of sloppy spec check) holds, but the solution of using feature flag was a tunnel vision from other discussions.

src/chain/optimism.rs
kien-rise: Sorry for the confusion.
- On Optimism "Testnet" (not sure if it is Optimism Sepolia or whatever): Regolith is activated after Bedrock.
- On Optimism Mainnet, Regolith and Bedrock are activated at the same time (block number = 105235063).

As a result, this fn should never return `SpecId::BEDROCK`.

------

Based on `reth` repo, the implementation should be:

```rust
    fn get_block_spec(&self, header: &Header) -> Result<SpecId, Self::BlockSpecError> {
        // TODO: The implementation below is only true for Optimism Mainnet.
        // When supporting other networks (e.g. Optimism Sepolia), remember to adjust the code here.
        if header.timestamp >= 1720627201 {
            Ok(SpecId::FJORD)
        } else if header.timestamp >= 1710374401 {
            Ok(SpecId::ECOTONE)
        } else if header.timestamp >= 1704992401 {
            Ok(SpecId::CANYON)
        } else if header.number >= 105235063 {
            Ok(SpecId::REGOLITH)
        } else if header.number >= 3950000 {
            Ok(SpecId::BERLIN)
        } else {
            Ok(SpecId::MUIR_GLACIER)
        }
    }
```

However, `revm` does not seem to support pre-Bedrock blocks (i.e. when block number < 105235063).

PS: This observation is the same as my previous observation a long time ago. Basically, I was just redoing whatever I have done previously.

src/vm.rs
kien-rise: > Wait, the actual fix is to add to Chain things like has_eip_161, has_eip_658, etc. that take in a SpecId. The problem (of sloppy spec check) holds, but the solution of using feature flag was a tunnel vision from other discussions.

Yes, that's what I mentioned in the last sentence of my last comment.

Feature flags (except OS, CPU related) are for optimization only, not for correctness. Anyone can compile with `--all-features` and expect the code to work.

kien-rise: > add to Chain things like has_eip_161, has_eip_658

shouldn't this be done in another PR?

src/chain/optimism.rs
kien-rise: Added some comments to explain.

src/vm.rs
hai-rise: The proper way is to have them added to `Chain`, replace the current checks, and merge it first. Then in this one we can focus on the implementation of OP, also as a clean reference of what a complete support looks like for new chains.

kien-rise: See https://github.com/risechain/pevm/pull/347

kien-rise: So, should the check be changed from:

```
if self.spec_id.is_enabled_in(SpecId::LONDON)
```

to 

```
if self.chain.is_eip_1559_enabled(self.spec_id)
```

?

----

Reference
https://eips.ethereum.org/EIPS/eip-1559

hai-rise: Yes

kien-rise: See https://github.com/risechain/pevm/pull/349

kien-rise: Resolved by rebasing.

PR #234, opened 2024-07-24
refactor: add get_reward_policy in trait PevmChain

kien-rise:

### DIFF
--- a/src/chain.rs
+++ b/src/chain.rs
@@ -11,6 +11,7 @@ use revm::{
 
 use crate::{
     mv_memory::{LazyAddresses, MvMemory},
+    vm::RewardPolicy,
     BuildIdentityHasher,
 };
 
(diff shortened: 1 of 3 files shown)


src/chain.rs
```diff
@@ -8,6 +8,7 @@ use revm::primitives::{BlockEnv, SpecId, TxEnv};
 
 use crate::{
     mv_memory::{LazyAddresses, MvMemory},
+    vm::RewardPolicy,
```
hai-falcony, line 11: Let's bring the `RewardPolicy` definition from `vm` to `chain`.

hai-rise, line 11: Let's bring `RewardPolicy` from `vm` to `chain`.

kien-rise: ok

Issue #202, opened 2024-07-04
Deduplicate bytecodes


hai-rise:
Many accounts share the same bytecode (forks) so it's much more efficient only to track the code hash accompanied by a mapping from code hashes to the bytecodes. Basically, we should deduplicate any container of `EvmAccount`.
- [ ] `EvmStateTransitions`
- [ ] `InMemoryStorage`
- [x] `./blocks`

PR #245, opened 2024-08-05
feat: deduplicate bytecodes in InMemoryStorage

kien-rise:
Partially resolves https://github.com/risechain/pevm/issues/202

### DIFF
--- a/src/storage/in_memory.rs
+++ b/src/storage/in_memory.rs
@@ -12,8 +12,7 @@ type Accounts = HashMap<Address, EvmAccount, BuildAddressHasher>;
 #[derive(Debug, Default, Clone)]
 pub struct InMemoryStorage<'a> {
     accounts: Accounts,
-    bytecodes_ref: Option<&'a AHashMap<B256, EvmCode>>,
-    bytecodes: AHashMap<B256, EvmCode>,
+    bytecodes: Option<&'a AHashMap<B256, EvmCode>>,
     block_hashes: AHashMap<u64, B256>,
 }
 
@@ -21,32 +20,7 @@ impl<'a> InMemoryStorage<'a> {
     /// Construct a new [InMemoryStorage]
     pub fn new(
         accounts: impl IntoIterator<Item = (Address, EvmAccount)>,
-        bytecodes: impl IntoIterator<Item = (B256, EvmCode)>,
-        block_hashes: impl IntoIterator<Item = (u64, B256)>,
-    ) -> Self {
-        let mut accounts: Accounts = accounts.into_iter().collect();
-        let mut bytecodes: AHashMap<B256, EvmCode> = bytecodes.into_iter().collect();
-        let block_hashes: AHashMap<u64, B256> = block_hashes.into_iter().collect();
-
-        for account in accounts.values_mut() {
-            let code = account.code.take();
-            if let (Some(code), Some(code_hash)) = (code, account.code_hash) {
-                bytecodes.insert(code_hash, code);
-            }
-        }
-
-        InMemoryStorage {
-            accounts,
-            bytecodes_ref: None,
-            bytecodes,
-            block_hashes,
-        }
-    }
-
-    /// Construct a new [InMemoryStorage] with `bytecodes_ref`
-    pub fn new_with_bytecodes_ref(
-        accounts: impl IntoIterator<Item = (Address, EvmAccount)>,
-        bytecodes_ref: &'a AHashMap<B256, EvmCode>,
+        bytecodes: Option<&'a AHashMap<B256, EvmCode>>,
         block_hashes: impl IntoIterator<Item = (u64, B256)>,
     ) -> Self {
         let accounts: Accounts = accounts.into_iter().collect();
@@ -58,8 +32,7 @@ impl<'a> InMemoryStorage<'a> {
 
         InMemoryStorage {
             accounts,
-            bytecodes_ref: Some(bytecodes_ref),
-            bytecodes: AHashMap::new(),
+            bytecodes,
             block_hashes,
         }
     }
@@ -84,12 +57,10 @@ impl<'a> Storage for InMemoryStorage<'a> {
     }
 
     fn code_by_hash(&self, code_hash: &B256) -> Result<Option<EvmCode>, Self::Error> {
-        if let Some(bytecodes) = self.bytecodes_ref {
-            if let Some(evm_code) = bytecodes.get(code_hash) {
-                return Ok(Some(evm_code.clone()));
-            }
-        }
-        Ok(self.bytecodes.get(code_hash).cloned())
+        Ok(match self.bytecodes {
+            Some(bytecodes) => bytecodes.get(code_hash).cloned(),
+            None => None,
+        })
     }
 
     fn has_storage(&self, address: &Address) -> Result<bool, Self::Error> {
--- a/src/storage/in_memory.rs
+++ b/src/storage/in_memory.rs
@@ -22,19 +22,24 @@ impl InMemoryStorage {
     // [accounts].
     pub fn new(
         accounts: impl IntoIterator<Item = (Address, EvmAccount)>,
+        bytecodes: impl IntoIterator<Item = (B256, EvmCode)>,
         block_hashes: impl IntoIterator<Item = (u64, B256)>,
     ) -> Self {
-        let accounts: Accounts = accounts.into_iter().collect();
-        let mut bytecodes = AHashMap::default();
-        for (_, account) in accounts.iter() {
-            if let (Some(code_hash), Some(code)) = (account.code_hash, &account.code) {
-                bytecodes.entry(code_hash).or_insert_with(|| code.clone());
+        let mut accounts: Accounts = accounts.into_iter().collect();
+        let mut bytecodes: AHashMap<B256, EvmCode> = bytecodes.into_iter().collect();
+        let block_hashes: AHashMap<u64, B256> = block_hashes.into_iter().collect();
+
+        for account in accounts.values_mut() {
+            let code = account.code.take();
+            if let (Some(code), Some(code_hash)) = (code, account.code_hash) {
+                bytecodes.insert(code_hash, code);
             }
         }
+
         InMemoryStorage {
             accounts,
             bytecodes,
-            block_hashes: block_hashes.into_iter().collect(),
+            block_hashes,
         }
     }
 }
--- a/src/storage/in_memory.rs
+++ b/src/storage/in_memory.rs
@@ -23,17 +23,10 @@ impl<'a> InMemoryStorage<'a> {
         bytecodes: Option<&'a AHashMap<B256, EvmCode>>,
         block_hashes: impl IntoIterator<Item = (u64, B256)>,
     ) -> Self {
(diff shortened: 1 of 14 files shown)


src/storage/rpc.rs
```diff
@@ -103,13 +105,24 @@ impl<N: Network> Storage for RpcStorage<N> {
...
+            }
             let basic = AccountBasic { balance, nonce };
             self.cache_accounts.lock().unwrap().insert(
                 *address,
                 EvmAccount {
                     basic: basic.clone(),
-                    code_hash: (!code.is_empty()).then(|| code.hash_slow()),
-                    code: (!code.is_empty()).then(|| code.into()),
+                    code_hash,
+                    code: None,
```
hai-rise, line 125: We now don't snapshot new codes?

src/storage/rpc.rs
```diff
@@ -103,13 +105,24 @@ impl<N: Network> Storage for RpcStorage<N> {
                 return Ok(None);
             }
             let code = Bytecode::new_raw(code);
+            let code_hash = if code.is_empty() {
+                None
+            } else {
+                Some(code.hash_slow())
+            };
+            if let Some(code_hash) = code_hash {
+                self.cache_bytecodes
```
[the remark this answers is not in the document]
kien-rise: 1. Ok
2. Can I rename the PR title to `feat: deduplicate bytecodes in storages` instead? The PR is quite small btw.

tests/common/mod.rs
```diff
@@ -87,6 +81,9 @@ pub fn for_each_block_from_disk(mut handler: impl FnMut(Block, InMemoryStorage))
                 })
                 .unwrap_or_default();
 
-        handler(block, InMemoryStorage::new(accounts, block_hashes));
+        handler(
+            block,
+            InMemoryStorage::new(accounts, bytecodes.clone(), block_hashes),
```
hai-rise, line 86: Not important now, but this `clone` suggests that `InMemoryStorage` can also reference external data like `bytecodes` instead of owning it for better performance.

src/storage/rpc.rs
kien-rise: good catch, we should add and use `pub fn get_cache_bytecodes`

src/storage/rpc.rs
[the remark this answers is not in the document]
kien-rise: ok, I will proceed with 1, which is also my prefered option

src/storage/rpc.rs
kien-rise: resolved in https://github.com/risechain/pevm/pull/248

src/storage/in_memory.rs
```diff
@@ -10,36 +10,62 @@ type Accounts = HashMap<Address, EvmAccount, BuildAddressHasher>;
 
 /// A storage that stores chain data in memory.
 #[derive(Debug, Default, Clone)]
-pub struct InMemoryStorage {
+pub struct InMemoryStorage<'a> {
     accounts: Accounts,
+    bytecodes_ref: Option<&'a AHashMap<B256, EvmCode>>,
```
hai-rise, line 15: Just remove the previous `bytecodes` and make this one `bytecodes`? For empty mock bytecodes just pass `None` in.

src/storage/in_memory.rs
```diff
@@ -10,36 +10,62 @@ type Accounts = HashMap<Address, EvmAccount, BuildAddressHasher>;
...
         InMemoryStorage {
             accounts,
+            bytecodes_ref: None,
             bytecodes,
-            block_hashes: block_hashes.into_iter().collect(),
+            block_hashes,
+        }
+    }
+
+    /// Construct a new [InMemoryStorage] with `bytecodes_ref`
```
hai-rise, line 46: We really don't want this extra complication.

src/storage/in_memory.rs
kien-rise: that makes perfect sense

src/storage/in_memory.rs
```diff
@@ -10,36 +10,35 @@ type Accounts = HashMap<Address, EvmAccount, BuildAddressHasher>;
...
         let accounts: Accounts = accounts.into_iter().collect();
-        let mut bytecodes = AHashMap::default();
-        for (_, account) in accounts.iter() {
-            if let (Some(code_hash), Some(code)) = (account.code_hash, &account.code) {
-                bytecodes.entry(code_hash).or_insert_with(|| code.clone());
-            }
+        let block_hashes: AHashMap<u64, B256> = block_hashes.into_iter().collect();
+
+        for account in accounts.values() {
+            assert!(account.code.is_none());
```
hai-rise, line 30: This is stricter than needed since these `code`s are unused, their actual values don't affect correctness. We should just inline the `collect`s and return an `InMemoryStorage` directly.

kien-rise: ok

tests/erc20/main.rs
```diff
@@ -9,17 +9,27 @@ pub mod common;
...
 use revm::primitives::{Address, TxEnv};
 
 #[test]
 fn erc20_independent() {
     const N: usize = 37123;
     let (mut state, txs) = generate_cluster(N, 1, 1);
     state.insert(Address::ZERO, EvmAccount::default()); // Beneficiary
-    test_execute_revm(InMemoryStorage::new(state, []), txs);
+
+    let mut bytecodes: AHashMap<B256, EvmCode> = AHashMap::new();
```
hai-rise, line 24: Let's return `bytecodes` from `generate_cluster`?

tests/uniswap/main.rs
```diff
@@ -29,5 +30,17 @@ fn uniswap_clusters() {
         final_state.extend(state);
         final_txs.extend(txs);
     }
-    common::test_execute_revm(InMemoryStorage::new(final_state, []), final_txs)
+
+    let mut bytecodes: AHashMap<B256, EvmCode> = AHashMap::new();
```
hai-rise, line 34: Let's return `bytecodes` from `generate_cluster`.

src/storage/in_memory.rs
kien-rise: Fixed.

Keep in mind that this `assert!` statement has shown its effectiveness in the recent CI failure: https://github.com/risechain/pevm/actions/runs/10263383045/job/28395093118?pr=245

Without this `assert!`, the error log will be unmeaningful:
```
---- ethereum_state_tests stdout ----
thread '<unnamed>' panicked at tests/ethereum/main.rs:219:21:
assertion `left == right` failed: Mismatched state root for "tests/ethereum/tests/GeneralStateTests/stRandom2/randomStatetest534.json"
  left: 0x0dc060d876225462f5269e64bd3bcea7b687d760181636f9d6b1805657b1121b
 right: 0x0746899e8148565c59dd0b63c61acafc2b1ac75aa977c3a620b2b3a9c0667a5f
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread '<unnamed>' panicked at tests/ethereum/main.rs:219:21:
assertion `left == right` failed: Mismatched state root for "tests/ethereum/tests/GeneralStateTests/stRevertTest/RevertInDelegateCall.json"
  left: 0xe6960d03cae79440ffbf8b3deeeebc75da0b69b812aed7170c05b5949816aea7
 right: 0xd30a9c46d689c5fead9ce54e1b2f441bde1ed776daf747d9de01df716bbbbe2c
```

tests/erc20/main.rs
kien-rise: resolved in the latest commit

tests/uniswap/main.rs
kien-rise: resolved in the latest commit

tests/common/mod.rs
```diff
@@ -49,7 +50,7 @@ pub const RAW_TRANSFER_GAS_LIMIT: u64 = 21_000;
 // TODO: Put somewhere better?
 pub fn for_each_block_from_disk(mut handler: impl FnMut(Block, InMemoryStorage)) {
     // Parse bytecodes
-    let bytecodes: HashMap<B256, EvmCode> = bincode::deserialize_from(BufReader::new(
+    let bytecodes: AHashMap<B256, EvmCode> = bincode::deserialize_from(BufReader::new(
```
hai-rise, line 53: Just use `Bytecodes`?

tests/common/mod.rs
```diff
@@ -14,6 +14,7 @@ pub use runner::{assert_execution_result, mock_account, test_execute_alloy, test
 pub mod storage;
 
 pub type ChainState = AHashMap<Address, EvmAccount>;
+pub type Bytecodes = AHashMap<B256, EvmCode>;
```
hai-rise, line 17: We can even move this to and use it in `InMemoryStorage`.

tests/erc20/mod.rs
```diff
@@ -61,5 +61,13 @@ pub fn generate_cluster(
         }
     }
 
-    (state, txs)
+    let mut bytecodes = Bytecodes::new();
+    for account in state.values_mut() {
```
hai-rise, line 65: This pattern is very unnatural. It is much better to declare `bytecodes` with `state` and `txs`, and make `ERC20Token` return `account` and `bytecode` separately. Or at least take the `code` from the returned account there instead of here.

kien-rise: basically you want to cherry pick this commit? https://github.com/risechain/pevm/pull/236/commits/66da2bc94c6e98086a67468b241688ea0f99cd2e

hai-rise: After looking at the Uniswap case this is much cleaner and less error-prone, and it's better to unify both instead of noisily change the `ERC20Token` interface. Let's just keep it as is.

tests/common/mod.rs
kien-rise: `AHashMap<B256, EvmCode>` is used in various places. For example, in `src/storage/rpc.rs`, we have `cache_bytecodes: Mutex<AHashMap<B256, EvmCode>>`.

Should we move the type definition to `src/lib.rs`?

src/storage/in_memory.rs
hai-rise: Pretty nice but not worth it. Also, if we aim for helpfulness then we should error when there are codes in accounts but `bytecodes` is `None`. Having codes in both seem like valid usage and wasn't the root cause of the failed test anyway.

tests/common/mod.rs
hai-rise: Yeah, let's do it.

hai-rise: Ah, we can put it in `storage` where `EvmCode` and the friends are. Not the best place but we can move everything together later.

tests/erc20/mod.rs
hai-rise: Removing `EvmAccount::code` is kinda nice since having `Some(code_hash)` and ` None` `code` is very confusing. Let's consider that in the next PR where we introduce the bytecodes map that accompany the accounts in the execution output. Changing the contract build interfaces can wait until then, too.

tests/common/mod.rs
kien-rise: Fixed in the latest commit.

PR #246, opened 2024-08-05
feat: add OnDiskStorage

kien-rise:


examples/to_mdbx.rs
```diff
@@ -0,0 +1,134 @@
...
+fn main() -> Result<(), Error> {
+    let args = Args::parse();
+    let tables = read_tables_from(args.bytecodes, args.pre_state)?;
+    let db = open_db(args.output).map_err(Error::other)?;
+    write_table_to(&db, "balance", tables.balance.iter()).map_err(Error::other)?;
+    write_table_to(&db, "nonce", tables.nonce.iter()).map_err(Error::other)?;
+    write_table_to(&db, "code_hash", tables.code_hash.iter()).map_err(Error::other)?;
+    write_table_to(&db, "code_by_hash", tables.code_by_hash.iter()).map_err(Error::other)?;
+    write_table_to(&db, "has_storage", tables.has_storage.iter()).map_err(Error::other)?;
+    write_table_to(&db, "storage", tables.storage.iter()).map_err(Error::other)?;
```
kien-rise, line 132: TODO: missing table `block_hash`

Cargo.toml
```diff
@@ -30,11 +30,13 @@ revm = { git = "https://github.com/risechain/revm", rev = "7b42abb672deacde9e053
 alloy-provider = "0.2.1"
 alloy-transport = "0.2.1"
 alloy-transport-http = "0.2.1"
+libmdbx = "0.5.0"
```
hai-falcony, line 33: Let's put this in its own `# On-Disk Storage dependencies` section with the same dedicated feature TODO.

hai-rise, line 33: Let's put this under its own `# On-Disk Storage dependencies` "section" with the same dedicated feature TODO.

examples/to_mdbx.rs
```diff
@@ -0,0 +1,150 @@
+//! Convert a block folder to a MDBX folder
```
hai-rise, line 1: This seems to be too important to be put under `examples`. Until we have a better structure it's easier to just put it in `tests`, to be imported in automated tests & benchmarks (not manually).

examples/to_mdbx.rs
```diff
@@ -0,0 +1,150 @@
...
+//! For help, run: `cargo run --example to_mdbx -- --help`
+
+use alloy_primitives::{Address, Bytes, FixedBytes, B256, B64, U64};
+use clap::Parser;
```
hai-rise, line 5: We don't want these complications this early. `data` is already hard-coded for tests & benchmarks, let's just do the same here. Practically there is little demand to run any of these manually.

examples/to_mdbx.rs
```diff
@@ -0,0 +1,150 @@
...
+    #[clap(long, value_name = "FILE")]
+    block_hashes: Option<String>,
+    /// Path to output MDBX dir
+    #[clap(long, value_name = "DIR")]
+    output: String,
+}
+
+#[derive(Debug, Clone, Default)]
+struct Tables {
+    balance: HashMap<Address, B256>,
```
hai-rise, line 42: Why don't we just map `Address` to a (serialized) `EvmAccount` on disk? This granularity suggests that we'd need many expensive I/O operations to fetch an account. State commitments will also be very slow. 

Ideally, we would use the least I/O then cache hot accounts in memory anyway.

hai-rise, line 42: Why don't we just map `Address` to a (serialized) `EvmAccount` on disk? This granularity suggests that we'd need many expensive I/O operations to fetch an account. State commitments will also be very slow. 

Ideally, we would use the least I/O then cache hot accounts in memory anyway.

examples/to_mdbx.rs
```diff
@@ -0,0 +1,150 @@
...
+    has_storage: HashMap<Address, B0>,
+    storage: HashMap<B416, B256>,
+    block_hash: HashMap<B64, B256>,
+}
+
+fn open_db(dir: impl AsRef<Path>) -> Result<Database<NoWriteMap>, libmdbx::Error> {
+    const MB: isize = 1048576;
+
+    #[allow(clippy::identity_op)]
+    Database::<NoWriteMap>::open_with_options(
```
hai-rise, line 55: Will `open` work for now? Ideally we'll have a dedicated task/PR for option tuning.

examples/to_mdbx.rs
```diff
@@ -0,0 +1,150 @@
...
+    let tx = db.begin_rw_txn()?;
+    let table = tx.create_table(Some(table_name), TableFlags::default())?;
+    for (k, v) in entries {
+        tx.put(&table, k, v, WriteFlags::UPSERT)?;
+    }
+    tx.commit()?;
+    Ok(())
+}
+
+fn main() -> Result<(), Error> {
```
hai-rise, line 137: This should be a simple function (with hard-coded paths) that:
1. Opens a database.
2. Reads and write accounts.
3. Reads and write bytecodes.
4. Reads and write block hashes.

Things like `Args` and `Tables` are unnecessary.

examples/run_on_mdbx.rs
```diff
@@ -0,0 +1,47 @@
...
+        let file = File::open(args.block)?;
+        serde_json::from_reader(BufReader::new(file))?
+    };
+
+    let on_disk_storage = OnDiskStorage::open(args.mdbx).map_err(Error::other)?;
+    let wrapped_storage = StorageWrapper(&on_disk_storage);
+    let db = CacheDB::new(&wrapped_storage);
+
+    let chain = PevmEthereum::mainnet();
+    common::test_execute_alloy(&db, &chain, block, true);
```
hai-rise, line 44: Let's port this to `tests`.

src/storage/on_disk.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+impl Storage for OnDiskStorage {
+    type Error = libmdbx::Error;
+
+    fn basic(
+        &self,
+        address: &alloy_primitives::Address,
+    ) -> Result<Option<super::AccountBasic>, Self::Error> {
+        let tx = self.inner.begin_ro_txn()?;
```
hai-rise, line 36: 1. Can we reuse the same read-only transaction (pool) instead of creating a new one on each read?
2. Is it more performant to run concurrent transactions in separate threads for parallel mode?

src/storage/on_disk.rs
```diff
@@ -0,0 +1,113 @@
...
+use alloy_primitives::{keccak256, Address, Bytes, FixedBytes, B256, B64, U256};
+use libmdbx::{Database, DatabaseOptions, NoWriteMap};
+use revm::primitives::Bytecode;
+
+use super::{AccountBasic, EvmCode, Storage};
+
+/// A storage that stores chain data in a MDBX database.
+#[derive(Debug)]
+pub struct OnDiskStorage {
+    inner: Database<NoWriteMap>,
```
hai-rise, line 12: `db` over `inner`?

src/storage/on_disk.rs
```diff
@@ -0,0 +1,113 @@
+use std::path::Path;
+
+use alloy_primitives::{keccak256, Address, Bytes, FixedBytes, B256, B64, U256};
+use libmdbx::{Database, DatabaseOptions, NoWriteMap};
+use revm::primitives::Bytecode;
+
+use super::{AccountBasic, EvmCode, Storage};
+
+/// A storage that stores chain data in a MDBX database.
```
hai-rise, line 9: `A storage that reads data from an on-disk MDBX database.`

src/storage/on_disk.rs
```diff
@@ -0,0 +1,113 @@
...
+#[derive(Debug)]
+pub struct OnDiskStorage {
+    inner: Database<NoWriteMap>,
+}
+
+impl OnDiskStorage {
+    /// Opens the on-disk storage at the specified path.
+    pub fn open(path: impl AsRef<Path>) -> Result<Self, libmdbx::Error> {
+        let db = Database::open_with_options(
```
hai-rise, line 18: Will `open` work for now?

src/storage/on_disk.rs
```diff
@@ -0,0 +1,113 @@
...
+impl Storage for OnDiskStorage {
+    type Error = libmdbx::Error;
+
+    fn basic(
+        &self,
+        address: &alloy_primitives::Address,
+    ) -> Result<Option<super::AccountBasic>, Self::Error> {
+        let tx = self.inner.begin_ro_txn()?;
+        let Some(balance) = tx
+            .open_table(Some("balance"))
```
hai-rise, line 38: Is it slow to call `open_table` on every read?

examples/to_mdbx.rs
kien-rise: `open` does not work, it will cause error `MDBX_MAP_FULL`.

By default, the max num tables is 1 (the default table), and the max db file size is a few of MBs (which is too small).

fyi, https://github.com/erthink/libmdbx/issues/136

examples/run_on_mdbx.rs
kien-rise: > Let's port this to an automated test.

Ok, no problem. How about making an automated test that for each block `K`:
1. Read from `data/blocks/<K>/*.json` and write to `data/blocks/<K>/mdbx.dat`
2. Run `pevm` using `OnDiskStorage`
3. Finally, delete `data/blocks/<K>/mdbx.dat`

?

src/storage/on_disk.rs
kien-rise: These are for performance optimization only, right? Can we postpone to another PR?

At least we should measure the current performance first. Otherwise, these optimizations look premature to me.

src/storage/on_disk.rs
kien-rise: No. We need to specify `max_tables`.

```
Err(ExecutionError("database error: MDBX_DBS_FULL: Too many DBI-handles (maxdbs reached)"))
```

src/storage/on_disk.rs
kien-rise: We need a benchmark to answer this question.

examples/to_mdbx.rs
hai-rise: I see, then let's put the options in a `const`/`static` with a comment on each custom config on why it's there.

examples/run_on_mdbx.rs
hai-rise: Yeah, we can add `OnDiskStorage` to `for_each_block_from_disk`? Then for each block we'll test & bench with both in memory and on disk storage.

Not important now, but eventually we can check in a DB file with all the bytecodes and block hashes as a base DB, to then load and drop accounts for each test/bench.

src/storage/on_disk.rs
hai-rise: We should benchmark and optimize right from this PR, so we don't have to check in and maintain unwanted code if it fails to demonstrate great speedup over in-memory storage. We can break the deduplicating bytecodes task down because each part is an instant improvement we want (to maintain).

These specific questions are important to improve our understanding. So that if on-disk storage shows promise but MDBX flops then hopefully we know why and seek a better alternative, like a more async I/O-friendly database and schema to avoid locks among worker threads.

So yeah, let's add a benchmark and optimize from there.

src/storage/on_disk.rs
hai-rise: We should take in the options then since the callee knows these limits the best. Hardcoding anything here will only reduce usage.

examples/run_on_mdbx.rs
kien-rise: > we can add OnDiskStorage to for_each_block_from_disk

We cannot do this because the function signature is:

```
pub fn for_each_block_from_disk(
  mut handler: impl FnMut(Block, InMemoryStorage)
)
```

instead of:

```
pub fn for_each_block_from_disk(
    mut handler: impl FnMut(Block, Box<dyn Storage<Error = dyn Display>>),
)
```

------

Also, looking at `benches/mainnet.rs`, if we call `handler` twice in `for_each_block_from_disk`, it seems that the second call will override all the output (i.e. `target/criterion/*`) from the first call.

kien-rise: My suggestion:

1. Keep `for_each_block_from_disk` as is.
2. Create a new test fn in `tests/mainnet.rs` and a new bench fn in `benches/mainnet.rs`

hai-rise: No, I meant for `for_each_block_from_disk` to return both an `InMemoryStorage` and an `OnDiskStorage`.

kien-rise: ok, that sounds doable

kien-rise: How about this idea?

```rust
struct SampledBlock {
  path_to_block: PathBuf,
  path_to_pre_state: PathBuf,
  path_to_block_hashes: Option<PathBuf>,
}

impl SampledBlock {
  fn get_block(&self) -> Result<Block, std::io::Error>;
  fn get_in_memory_storage(&self, bytecodes: &Bytecodes) -> Result<InMemoryStorage, std::io::Error>;
  fn get_on_disk_storage(&self, bytecodes: &Bytecodes) -> Result<OnDiskStorage, std::io::Error>;
}

fn get_all_sampled_blocks() -> Result<Vec<SampledBlock>, std::io::Error>;
fn get_bytecodes() -> Result<Bytecodes, std::io::Error>;
```

-----

With this struct, the callees can do like:

```rust
let bytecodes = get_bytecodes()?;
for sample in get_all_sampled_blocks()? {
  let block = sample.get_block()?;
  let in_memory_storage = sample.get_in_memory_storage(&bytecodes)?;
  let on_disk_storage = sample.get_on_disk_storage(&bytecodes)?;
  ...
}
```

hai-rise: No, we need quick iterative benchmarks to focus on storage optimization. Stop adding code & abstractions that solve (literally) nothing while adding noise to development, code review, and future maintenance costs.

src/storage/on_disk.rs
kien-rise: > the callee knows these limits the best

`OnDiskStorage` is assuming that the table names are `balance`, `nonce`, `code_hash`, etc. So I think it is safe to assume that `OnDiskStorage` already know what tables there are and how many tables there are. So I think `OnDiskStorage` should dictate the param `max_tables` instead of letting callees specifying it.

examples/run_on_mdbx.rs
kien-rise: ok, understood

Cargo.toml
```diff
@@ -33,8 +33,12 @@ alloy-transport-http = "0.2.1"
 reqwest = "0.12.5"
 tokio = { version = "1.39.2", features = ["rt-multi-thread"] }
 
-[dev-dependencies]
+# MDBX Storage dependencies
+libmdbx = "0.5.0"
 bincode = "1.3.3"
+
+[dev-dependencies]
+clap = { version = "4.5.13", features = ["derive", "env"] }
```
hai-rise, line 41: We don't need this anymore?

benches/mainnet.rs
```diff
@@ -25,35 +25,61 @@ pub fn criterion_benchmark(c: &mut Criterion) {
...
-    common::for_each_block_from_disk(|block, storage| {
+    common::for_each_block_from_disk(|block, in_memory_storage, on_disk_storage| {
         let mut group = c.benchmark_group(format!(
             "Block {}({} txs, {} gas)",
             block.header.number.unwrap(),
             block.transactions.len(),
             block.header.gas_used
         ));
-        group.bench_function("Sequential", |b| {
+        group.bench_function("Sequential (in memory)", |b| {
```
hai-rise, line 35: `/In Memory` to match `criterion`'s current format better.

benches/mainnet.rs
```diff
@@ -25,35 +25,61 @@ pub fn criterion_benchmark(c: &mut Criterion) {
...
+                    black_box(&in_memory_storage),
                     black_box(&chain),
                     black_box(block.clone()),
                     black_box(concurrency_level),
                     black_box(true),
                 )
             })
         });
-        group.bench_function("Parallel", |b| {
+        group.bench_function("Parallel (in memory)", |b| {
```
hai-rise, line 46: `/In Memory` to match `criterion`'s current format better.

benches/mainnet.rs
```diff
@@ -25,35 +25,61 @@ pub fn criterion_benchmark(c: &mut Criterion) {
...
-                    black_box(&storage),
+                    black_box(&in_memory_storage),
                     black_box(&chain),
                     black_box(block.clone()),
                     black_box(concurrency_level),
                     black_box(false),
                 )
             })
         });
+        group.bench_function("Sequential (on_disk)", |b| {
```
hai-rise, line 57: `/On Disk` to match `criterion`'s current format better.

benches/mainnet.rs
```diff
@@ -25,35 +25,61 @@ pub fn criterion_benchmark(c: &mut Criterion) {
...
+                    black_box(&chain),
+                    black_box(block.clone()),
+                    black_box(concurrency_level),
+                    black_box(true),
+                )
+            })
+        });
+        on_disk_storage.clear_cache();
+
+        group.bench_function("Parallel (on_disk)", |b| {
```
hai-rise, line 70: `/On Disk` to match `criterion`'s current format better.

benches/mainnet.rs
```diff
@@ -25,35 +25,61 @@ pub fn criterion_benchmark(c: &mut Criterion) {
...
+                pevm::execute(
+                    black_box(&on_disk_storage),
+                    black_box(&chain),
+                    black_box(block.clone()),
+                    black_box(concurrency_level),
+                    black_box(true),
+                )
+            })
+        });
+        on_disk_storage.clear_cache();
```
hai-rise, line 68: Intuitively this should be put inside `b.iter` which runs `pevm:execute` several times to measure the average, etc. execution time. Otherwise only the first run will read from disk.

benches/mainnet.rs
```diff
@@ -25,35 +25,61 @@ pub fn criterion_benchmark(c: &mut Criterion) {
...
+                pevm::execute(
+                    black_box(&on_disk_storage),
+                    black_box(&chain),
+                    black_box(block.clone()),
+                    black_box(concurrency_level),
+                    black_box(false),
+                )
+            })
+        });
+        on_disk_storage.clear_cache();
```
hai-rise, line 81: Intuitively this should be put inside `b.iter` which runs `pevm:execute` several times to measure the average, etc. execution time. Otherwise only the first run will read from disk.

We may need to `iter_custom` for even more precision, to not count cache clearing time & result dropping time into execution time.

src/storage/on_disk.rs
```diff
@@ -0,0 +1,242 @@
...
+    TableFlags, WriteFlags,
+};
+
+use super::{AccountBasic, EvmAccount, EvmCode, Storage};
+
+/// A storage that reads data from an on-disk MDBX database.
+#[derive(Debug)]
+pub struct OnDiskStorage {
+    db: Database<NoWriteMap>,
+    cache_accounts: Mutex<AHashMap<Address, Option<EvmAccount>>>,
```
hai-rise, line 19: This is obviously bad for parallel as two threads reading different addresses still block on this single `Mutex`. Let's use `dashmap` for concurrent maps like in `MvMemory`.

src/storage/on_disk.rs
hai-rise: No. A `trait` requires the implementer to has some signatures but makes no assumption about the actual `struct`ure or implementation. Similarly, `OnDiskMemory` only has some requirements on some tables, but generally  `db: Database<NoWriteMap>` accepts any database schema. We cannot assume that, say, an Reth ExEx that uses `pevm` would have fewer tables, or force them to use different databases for different things.

src/storage/on_disk.rs
```diff
@@ -0,0 +1,242 @@
...
+}
+
+/// Create a temp dir containing MDBX
+pub fn create_db_dir<'a>(
+    block_number: &str,
+    bytecodes: impl Iterator<Item = (&'a B256, &'a EvmCode)>,
+    pre_state: impl Iterator<Item = (&'a Address, &'a EvmAccount)>,
+    block_hashes: impl Iterator<Item = (&'a u64, &'a B256)>,
+) -> Result<PathBuf, libmdbx::Error> {
+    let mut dir = std::env::temp_dir();
```
hai-rise, line 215: Why are these defined in library source instead of `tests/common`? This usage seems too niche to be useful for users. And we can always extract it back from tests when there is a clear proposed use case.

src/storage/on_disk.rs
```diff
@@ -0,0 +1,242 @@
...
+        "block_hashes",
+        block_hashes
+            .map(|(block_number, block_hash)| (Into::<B64>::into(*block_number), block_hash)),
+    )?;
+
+    Ok(dir)
+}
+
+/// Remove a temp dir
+pub fn remove_db_dir(db_dir: PathBuf) -> Result<(), std::io::Error> {
```
hai-rise, line 240: This thin wrapper is so redundant. How does `remove_db_dir(db_dir)` add anything to `remove_dir_all(&db_dir)`?

tests/common/runner.rs
```diff
@@ -53,10 +53,7 @@ pub fn test_execute_revm<S: Storage + Clone + Send + Sync>(storage: S, txs: Vec<
 
 // Execute an Alloy block sequentially & with PEVM and assert that
 // the execution results match.
-pub fn test_execute_alloy<
-    S: Storage + Clone + Send + Sync,
-    C: PevmChain + Send + Sync + PartialEq,
->(
+pub fn test_execute_alloy<S: Storage + Send + Sync, C: PevmChain + Send + Sync + PartialEq>(
```
hai-rise, line 56: Good catch. We can also make `block` a reference since we're going to `clone` below anyway.

src/storage/on_disk.rs
```diff
@@ -0,0 +1,242 @@
...
+            cache_accounts: Mutex::default(),
+            cache_bytecodes: Mutex::default(),
+            cache_block_hashes: Mutex::default(),
+        })
+    }
+
+    fn load_account_and_then<T, F: Fn(&Option<EvmAccount>) -> Result<T, libmdbx::Error>>(
+        &self,
+        address: &Address,
+        op: F,
```
hai-rise, line 45: Can we not just return an `EvmAccount` here?

src/storage/on_disk.rs
hai-rise: It's pretty big so we can just put it in `tests/common/mdbx.rs`.

src/storage/on_disk.rs
hai-rise: Also, why is the inner `Option`?

src/storage/on_disk.rs
kien-rise: Because `EvmAccount` is heavy, and we cannot return `&EvmAccount`.

hai-rise: I meant why do we go through `op` instead of just returning directly. And yes, returning a reference is good.

kien-rise: I get this compile error when trying to return a `&EvmAccount`:

```
cannot return value referencing local variable `evm_account`
returns a value referencing data owned by the current
```

I am checking if dashmap's `Ref` give me any possibilities.

src/storage/on_disk.rs
kien-rise: ok, let's just remove this fn

src/storage/on_disk.rs
hai-rise: Regardless of `DashMap` or `HashMap`, it should be possible by pushing `evm_account` into the cache then returning a reference into the cache.

src/storage/on_disk.rs
kien-rise: > why is the inner Option

If there is no account associated with an address, I will store `None` in the cache so that when the same query is performed, the cached value (`None`) will be returned immediately (instead of querying the `db`).

src/storage/on_disk.rs
kien-rise: ok, now it's returning `OccupiedEntry<Address, Option<EvmAccount>>`

hai-rise:
This is generally in the right direction, but closing because it's too old now, redoing it will be better than rebasing.
We should also pull in the `libmdbx` binding from Cargo instead of checking it in here.

PR #257, opened 2024-08-13
feat: deduplicate bytecodes in PevmResult

kien-rise:
Resolves https://github.com/risechain/pevm/issues/202

### DIFF
--- a/src/pevm.rs
+++ b/src/pevm.rs
@@ -5,8 +5,8 @@ use std::{
     thread,
 };
 
-use ahash::AHashMap;
-use alloy_primitives::U256;
+use ahash::{AHashMap, AHashSet};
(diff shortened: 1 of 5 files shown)


hai-rise (changes requested):
This PR is taking a wrong direction with expensive post-processing `clone`s and `collect`s. There is of course no speedup when we add ~constant cost (sequential post-processing) to both scenarios.

For improvement, we must separate bytecodes right from the `revm::ResultAndState` to `PevmTxExecutionResult` conversion. More specifically, we won't insert an `EvmCode` into the output `EvmAccount` but to a hash map if it doesn't already have the code hash.

This work is supposed to make both sequential & parallel faster.

PR #262, opened 2024-08-21
perf: fine tune concurrency level for mainnet benchmark

kien-rise:


kien-rise:
```
Average: x1.93
Max: x4.37
Min: x0.89
```

hai-rise:
We'll have more tasks like multiproof, maybe even garbage collecting, so we'll tune the concurrency level very differently in the next Parallel EVM.

PR #265, opened 2024-08-21
refactor: use fn push_origin to simplify code

jackwener:
use fn push_origin to simplify code

### DIFF
--- a/src/vm.rs
+++ b/src/vm.rs
@@ -166,10 +166,24 @@ impl<'a, S: Storage, C: PevmChain> VmDb<'a, S, C> {
         }
     }
 
+    // a function to push read_origins
+    fn push_origin(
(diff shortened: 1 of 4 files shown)


src/vm.rs
```diff
@@ -166,10 +166,24 @@ impl<'a, S: Storage, C: PevmChain> VmDb<'a, S, C> {
         }
     }
 
+    // a function to push read_origins
```
hai-rise, line 169: Let's elaborate a bit!
```
// Push a new read origin. Return an error when there's already 
// an origin but doesn't match the new one to force re-execution.
```

hai-rise (changes requested):
Minor comments but overall really like this simplification. Thank you!

jackwener:
@hai-rise Thank you very much for your careful review!

hai-rise (approved):
🫡

PR #266, opened 2024-08-22
feat: use AtomicUsize to replace Mutex<TxStatus>

kien-rise:


kien-rise:
Let me support the use cases where lock is not needed.

hai-rise:
We'll extend the transaction status in the next design, at least with the commit concept in BlockSTM v2 to commit batches of transactions as shred:
https://github.com/aptos-labs/aptos-core/blob/659530d327f640940977bc9d92e866a2dae8b5a3/aptos-move/block-executor/src/scheduler.rs#L141-L165

So let's keep things simple for now, can consider these micro-optimisatons once the new structure is in place.

PR #267, opened 2024-08-22
perf: use UnsafeVec to replace Vec<Mutex<>>

jackwener:
`last_locations` will only be updated in `record()`, `validate_read_locations()`, and `convert_writes_to_estimates()`. For the same `TxIdx`, these functions cannot execute concurrently, so we can replace `Vec<Mutex<>>` with `UnsafeVec`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -220,3 +220,4 @@ pub use storage::{
 };
 mod vm;
 pub use vm::{ExecutionError, PevmTxExecutionResult};
+mod unsafe_vec;
--- a/src/mv_memory.rs
+++ b/src/mv_memory.rs
@@ -5,8 +5,8 @@ use alloy_primitives::Address;
 use dashmap::{mapref::one::Ref, DashMap};
 
 use crate::{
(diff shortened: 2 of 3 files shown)


hai-rise:
@jackwener Nice insight as always! 

Actually, there can be concurrent validation tasks for the same transaction index.
There can only be one execution task when the status transitions from `ReadyToExecute` to `Executing`:
https://github.com/risechain/pevm/blob/cce4fc9216017d0e216cc3278396956e26d9a192/src/scheduler.rs#L125-L131

But a new validation task doesn't transition state, so multiple can be scheduled for the same executed/validated transaction:
https://github.com/risechain/pevm/blob/cce4fc9216017d0e216cc3278396956e26d9a192/src/scheduler.rs#L133-L140

We don't want a blocking `Validating` status as we want the latter task to start (with the latest data) as soon as possible. A relevant optimization I want to do is to have a way to signal the on-going validation task(s) to exit when a new one (of the transaction) starts.

Anyway, that is just extra information, as it should still be safe to have concurrent reads. 
I had a local branch but with `Vec<SyncUnsafeCell<LastLocations>>`, which is much shorter than rolling our own `UnsafeVec`. Do you think it's worth re-trying it that way?

PR #280, opened 2024-08-25
chore: remove useless `min()`

jackwener:
- remove useless `min()` 
- use debug_assert_eq to replace if/else

### DIFF
--- a/src/scheduler.rs
+++ b/src/scheduler.rs
@@ -209,61 +209,58 @@ impl Scheduler {
         next_validation_idx: TxIdx,
     ) -> Option<Task> {
         let mut tx = index_mutex!(self.transactions_status, tx_version.tx_idx);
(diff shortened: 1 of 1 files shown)


jackwener:
it's easy to review by seeing https://github.com/risechain/pevm/pull/280/files?diff=split&w=1

hai-rise:
The `min` is actually needed as `fetch_min` only returns the previous value. If it is larger than `next_validation_idx` then we want the local `min_validation_idx` to be ``next_validation_idx` immediately. Removing it apparently caused a deadlock in CI :pray:.

jackwener:
> 
The min is actually needed as fetch_min only returns the previous value. If it is larger than next_validation_idx then we want the local min_validation_idx to be next_validation_idx immediately. Removing it apparently caused a deadlock in CI 🙏.

Sorry, it was my oversight....

PR #289, opened 2024-08-28
feat: handle ERC20 tranfers lazily

kien-rise:


src/mv_memory.rs
```diff
@@ -35,8 +33,8 @@ pub struct MvMemory {
     pub(crate) data: DashMap<MemoryLocationHash, BTreeMap<TxIdx, MemoryEntry>, BuildIdentityHasher>,
     /// Last read & written locations of each transaction
     last_locations: Vec<Mutex<LastLocations>>,
-    /// Lazy addresses that need full evaluation at the end of the block
-    lazy_addresses: Mutex<LazyAddresses>,
+    /// Lazy locations that need full evaluation at the end of the block
+    lazy_locations: Mutex<HashSet<MemoryLocation, BuildSuffixHasher>>,
```
hai-rise, line 37: 1. Now that we have more lazy locations `DashSet` may be faster than `Mutex<HashSet<..>>` now. Worth trying a bench.
2. Using `BuildSuffixHasher` for `MemoryLocation` is pretty risky since not every possible value is an existing hash (of the same input distribution). I'd just use `ahash` for now.

src/mv_memory.rs
```diff
@@ -184,7 +185,7 @@ impl MvMemory {
         }
     }
 
-    pub(crate) fn consume_lazy_addresses(&self) -> impl IntoIterator<Item = Address> {
-        std::mem::take(&mut *self.lazy_addresses.lock().unwrap()).into_iter()
+    pub(crate) fn consume_lazy_locations(&self) -> impl IntoIterator<Item = MemoryLocation> {
+        std::mem::take(&mut *self.lazy_locations.lock().unwrap()).into_iter()
```
hai-rise, line 189: It may be faster to just return references here, since:
- We're cloning in post-processing anyway.
- This will be defer-dropped so pre-cleaning it may add latency instead of saving it.

src/pevm.rs
```diff
@@ -231,99 +231,136 @@ impl Pevm {
...
-                if !matches!(
-                    write_history.first_key_value(),
-                    Some((_, MemoryEntry::Data(_, MemoryValue::Basic(_))))
-                ) {
-                    if let Ok(Some(account)) = storage.basic(&address) {
-                        balance = account.balance;
-                        nonce = account.nonce;
+        for memory_location in mv_memory.consume_lazy_locations() {
+            let location_hash = self.hasher.hash_one(memory_location.clone());
+            let Some(write_history) = mv_memory.data.get(&location_hash) else {
```
hai-rise, line 236: This is good but the change is too noisy for this non-trivial work. I'd suggest doing it in another PR.

src/pevm.rs
```diff
@@ -231,99 +231,136 @@ impl Pevm {
...
+                                code_hash,
+                                code: code.clone(),
+                                storage: AHashMap::default(),
+                            });
                         }
-                        // TODO: Better error handling
-                        _ => unreachable!(),
                     }
+                }
+                MemoryLocation::CodeHash(_) => unreachable!(),
```
hai-rise, line 336: Minor but let's put the unreachable at the bottom.

src/vm.rs
```diff
@@ -90,6 +91,42 @@ pub(crate) enum VmExecutionResult {
...
+    },
+}
+
+impl LazyStrategy {
+    fn from(tx_sender: &Address, tx_recipient_code_hash: &Option<B256>, input: &Bytes) -> Self {
+        if tx_recipient_code_hash.is_none() {
+            return LazyStrategy::RawTransfer;
+        };
+
+        // TODO: We cannot blindly trust method_id. Use a whitelist of tx_recipient_code_hash.
```
hai-rise, line 117: Yeah, this should be configurable by users. Some use cases (that don't take external transactions) can trust all, some can whitelist, some can just turn off lazy ERC-20.

src/vm.rs
```diff
@@ -131,23 +166,28 @@ impl<'a, S: Storage, C: PevmChain> VmDb<'a, S, C> {
...
-        // locations with only one entry is much costlier than fully
-        // evaluating it concurrently.
         // TODO: Only lazy update in block syncing mode, not for block
         // building.
         if let Some(to) = to {
             db.to_code_hash = db.get_code_hash(*to)?;
-            db.is_lazy = db.to_code_hash.is_none()
-                && (vm.mv_memory.data.contains_key(&from_hash)
-                    || vm.mv_memory.data.contains_key(&to_hash.unwrap()));
+            db.lazy_strategy = LazyStrategy::from(from, &db.to_code_hash, &vm.txs[*tx_idx].data);
```
hai-rise, line 179: At this point we can just pass in `&tx` to cover `nonce`, from`, `to`, and `input`?

src/vm.rs
```diff
@@ -90,6 +91,42 @@ pub(crate) enum VmExecutionResult {
...
+    buf[12..32].copy_from_slice(address.as_slice());
+    keccak256(buf).into()
+}
+
+#[derive(Debug)]
+enum LazyStrategy {
+    None,
+    RawTransfer,
+    ERC20Transfer {
+        sender_balance_slot: U256,
```
hai-rise, line 106: Ideally we can simplify this as two 256-bit hashes are fairly large. This would make even a `LazyStrategy::None` 72 bytes in the stack.

src/vm.rs
```diff
@@ -222,7 +262,7 @@ impl<'a, S: Storage, C: PevmChain> Database for VmDb<'a, S, C> {
 
         // We return a mock for non-contract addresses (for lazy updates) to avoid
         // unnecessarily evaluating its balance here.
-        if self.is_lazy {
+        if matches!(self.lazy_strategy, LazyStrategy::RawTransfer) {
```
hai-rise, line 265: `==` is sufficient?

src/vm.rs
```diff
@@ -414,34 +454,75 @@ impl<'a, S: Storage, C: PevmChain> Database for VmDb<'a, S, C> {
     }
 
     fn storage(&mut self, address: Address, index: U256) -> Result<U256, Self::Error> {
+        if let LazyStrategy::ERC20Transfer {
+            sender_balance_slot,
+            recipient_balance_slot,
+        } = self.lazy_strategy
+        {
+            if index == sender_balance_slot {
+                return self
```
hai-rise, line 463: Can we mock `U256::MAX` here to save a storage read?

src/vm.rs
```diff
@@ -414,34 +454,75 @@ impl<'a, S: Storage, C: PevmChain> Database for VmDb<'a, S, C> {
...
+                    .vm
+                    .storage
+                    .storage(&address, &index)
+                    .map_err(|err| ReadError::StorageError(err.to_string()));
+            }
+            if index == recipient_balance_slot {
+                return self
```
hai-rise, line 470: Can we mock `U256::ZERO` here to save a storage read?

src/vm.rs
```diff
@@ -414,34 +454,75 @@ impl<'a, S: Storage, C: PevmChain> Database for VmDb<'a, S, C> {
...
+            }
+        }
+
         let location_hash = self
             .vm
             .hasher
             .hash_one(MemoryLocation::Storage(address, index));
 
         let read_origins = self.read_set.entry(location_hash).or_default();
+        read_origins.clear();
```
hai-rise, line 484: Let's at least add a `TODO` to proper handle read origins to avoid race conditions.

src/vm.rs
```diff
@@ -414,34 +454,75 @@ impl<'a, S: Storage, C: PevmChain> Database for VmDb<'a, S, C> {
...
+
+        let mut accumulated_addition = U256::ZERO;
```
hai-rise, line 486: Hmm, ideally we would put these logic in the lazy recipient branch above since unlike balance & nonce the operations on storage slots can be arbitrary (not addition in all cases).

src/vm.rs
hai-rise: Actually I think calling `push_origin` is pretty straightforward and can be done right away?

src/vm.rs
```diff
@@ -586,26 +667,31 @@ impl<'a, S: Storage, C: PevmChain> Vm<'a, S, C> {
...
+                                            ),
+                                        ));
+                                    } else if Some(account_location_hash) == to_hash {
+                                        write_set.push((
+                                            account_location_hash,
+                                            MemoryValue::LazyRecipient(tx.value),
+                                        ));
+                                    }
+                                }
+                                LazyStrategy::ERC20Transfer { .. } | LazyStrategy::None => {
```
hai-rise, line 682: A lazy ERC-20 transfer with zero ETH value can skip a write to the recipient basic here? Perhaps we can also apply the same zero ETH value check to raw transfer. 

We can also `LazySender` the sender here? I.e., we can mock the sender basic if it's interacting with ERC-20, too.

src/vm.rs
```diff
@@ -623,12 +705,43 @@ impl<'a, S: Storage, C: PevmChain> Vm<'a, S, C> {
...
+                        let memory_location = MemoryLocation::Storage(*address, slot);
+                        let memory_location_hash = self.hasher.hash_one(memory_location.clone());
+                        match evm.db().lazy_strategy {
+                            LazyStrategy::None => {
+                                write_set.push((
+                                    memory_location_hash,
+                                    MemoryValue::Storage(value.present_value),
+                                ));
+                            }
+                            LazyStrategy::RawTransfer => unreachable!(),
```
hai-rise, line 718: Minor but let's put all unreachables at the bottom.

src/vm.rs
```diff
@@ -623,12 +705,43 @@ impl<'a, S: Storage, C: PevmChain> Vm<'a, S, C> {
...
+                                        lazy_locations.push(memory_location);
+                                        write_set.push((
+                                            memory_location_hash,
+                                            MemoryValue::ERC20TransferSender(
+                                                value.original_value - value.present_value,
+                                            ),
+                                        ));
+                                    }
+                                    std::cmp::Ordering::Equal => {
+                                        unreachable!();
```
hai-rise, line 731: Minor but let's put all unreachables at the bottom.

src/vm.rs
```diff
@@ -623,12 +705,43 @@ impl<'a, S: Storage, C: PevmChain> Vm<'a, S, C> {
...
+                                    }
+                                    std::cmp::Ordering::Greater => {
+                                        lazy_locations.push(memory_location);
+                                        write_set.push((
+                                            memory_location_hash,
+                                            MemoryValue::ERC20TransferRecipient(
```
hai-rise, line 737: Looks like we only `match` here?

src/vm.rs
```diff
@@ -640,9 +753,7 @@ impl<'a, S: Storage, C: PevmChain> Vm<'a, S, C> {
 
                 drop(evm); // release db
 
-                if db.is_lazy {
-                    self.mv_memory.add_lazy_addresses([*from, *to.unwrap()]);
-                }
+                self.mv_memory.add_lazy_locations(lazy_locations);
```
hai-rise, line 756: Make sure either call site or the implementation checks for an empty array to avoid a meaningless lock.

src/pevm.rs
```diff
@@ -231,99 +231,136 @@ impl Pevm {
...
+                MemoryLocation::Storage(address, index) => {
+                    let mut storage_value = U256::ZERO;
+                    if let Ok(value) = storage.storage(&address, &index) {
```
hai-rise, line 339: We can use the same micro-optimization like in basic here: do not read from storage if the first write is an absolute value.

src/vm.rs
```diff
@@ -131,23 +170,28 @@ impl<'a, S: Storage, C: PevmChain> VmDb<'a, S, C> {
...
-        // evaluating it concurrently.
         // TODO: Only lazy update in block syncing mode, not for block
         // building.
         if let Some(to) = to {
             db.to_code_hash = db.get_code_hash(*to)?;
-            db.is_lazy = db.to_code_hash.is_none()
-                && (vm.mv_memory.data.contains_key(&from_hash)
-                    || vm.mv_memory.data.contains_key(&to_hash.unwrap()));
+            db.lazy_strategy = LazyStrategy::from(&db.to_code_hash, &vm.txs[*tx_idx].data);
+            if matches!(db.lazy_strategy, LazyStrategy::RawTransfer) {
```
hai-rise, line 184: `==` is enough?

src/mv_memory.rs
kien-rise: I am trying `DashSet`.

src/pevm.rs
kien-rise: I reverted the change. However, the added indentation is inevitable.

src/vm.rs
kien-rise: I added `PevmChain::is_erc20_transfer`:

```
/// Check if a tx is ERC20 transfer
fn is_erc20_transfer(&self, tx: &TxEnv) -> bool;
```

If a user wants to customize the logic, they can construct a `PevmChain` by themself.

src/vm.rs
```diff
@@ -90,6 +91,42 @@ pub(crate) enum VmExecutionResult {
     },
 }
 
+// https://docs.soliditylang.org/en/v0.8.26/internals/layout_in_storage.html#mappings-and-dynamic-arrays
+fn get_erc20_balance_slot(address: Address) -> U256 {
```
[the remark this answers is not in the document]
kien-rise: Thankfully I found a way to mitigate this. 

No more keccak256.

src/vm.rs
kien-rise: how the definition is `ERC20Transfer { amount: U256 },`

src/vm.rs
kien-rise: I have copy-pasted the logic from `basic`. No more race conditions.

The current handling is quite messy though.

src/vm.rs
kien-rise: Short answer: no.

Long answer: It's a bit hard to explain. The main idea is to perserve the zeroness of the storage slots.

Imagine Alice sends to Bob 10 tokens. Naturally, we think about `Alice -= 10; Bob += 10;`.
However, the following initial conditions must be treated differently:

1. Alice = 10. Bob = 0.
2. Alice = 11. Bob = 0.
3. Alice = 10. Bob = 1.
4. Alice = 11. Bob = 1.

Look at the zeroness, we have:

1. Alice: some -> none. Bob: none -> some.
2. Alice: some -> some. Bob: none -> some.
3. Alice: some -> none. Bob: some -> some.
4. Alice: some -> some. Bob: some -> some.

The zeroness affects the gas used.

We must choose one among 4 cases. If we choose wrongly, it is _very hard_ to fix. One might think about adjusting the beneficiary rewards in post. However, it's much more complicated.

Currently, at the end, if it is detected that we messed up with the zeroness, then it will fallback to sequential :(

For now, I claim that using storage value gives us the best prediction. At least for the current blocks, this strategy succeeds 99.99% of the times.

src/vm.rs
kien-rise: good idea!

src/vm.rs
kien-rise: 👍 I have added a comment to clarify:

```
Ordering::Equal => {
    // Unreachable because we are iterating `account.CHANGED_storage_slots()`.
    unreachable!();
}
```

src/vm.rs
kien-rise: added `let mut accumulated_subtraction = U256::ZERO; `

src/mv_memory.rs
kien-rise: In the latest benchmark, `DashSet` causes the performance to drop from 1.91 to 1.90.

So, it's inconclusive to say which one is better. Personally, using`DashSet` feels more comfortable.

kien-rise:
Closed. Moved to https://github.com/risechain/pevm/pull/346.

PR #332, opened 2024-09-05
feat: resolve error 429 Too Many Requests in CI

kien-rise:
Resolves https://github.com/risechain/pevm/pull/233#discussion_r1745015548

Community PRs are failing CI. It looks like `https://eth.llamarpc.com` does not work anymore (at least in our CI). This PR changes the default value to `https://eth.public-rpc.com`.

This PR also changes the env variable from `RPC_URL` to `ETHEREUM_RPC_URL`. Later we will have `OPTIMISM_RPC_URL`. Note that when the CI runs `cargo test`, both Ethereum and Optimism tests will run, therefore, having one `RPC_URL` is bad.

### DIFF
--- a/tests/mainnet.rs
+++ b/tests/mainnet.rs
@@ -21,7 +21,7 @@ pub mod common;
 // TODO: [tokio::test]?
 #[test]
 fn mainnet_blocks_from_rpc() {
-    let rpc_url = match std::env::var("RPC_URL") {
+    let rpc_url = match std::env::var("ETHEREUM_RPC_URL") {
         // The empty check is for GitHub Actions where the variable is set with an empty string when unset!?
         Ok(value) if !value.is_empty() => value.parse().unwrap(),
         _ => Url::parse("https://eth.public-rpc.com").unwrap(),
--- a/tests/mainnet.rs
+++ b/tests/mainnet.rs
@@ -24,7 +24,7 @@ fn mainnet_blocks_from_rpc() {
     let rpc_url = match std::env::var("RPC_URL") {
         // The empty check is for GitHub Actions where the variable is set with an empty string when unset!?
         Ok(value) if !value.is_empty() => value.parse().unwrap(),
-        _ => Url::parse("https://eth.llamarpc.com").unwrap(),
+        _ => Url::parse("https://eth.public-rpc.com").unwrap(),
     };
 
     // First block under 50 transactions of each EVM-spec-changing fork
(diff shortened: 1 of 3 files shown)


tests/mainnet.rs
```diff
@@ -21,10 +21,10 @@ pub mod common;
 // TODO: [tokio::test]?
 #[test]
 fn mainnet_blocks_from_rpc() {
-    let rpc_url = match std::env::var("RPC_URL") {
+    let rpc_url = match std::env::var("ETHEREUM_RPC_URL") {
```
hai-rise, line 24: We should also change `ci.yml`.

kien-rise: Thanks. I have fixed it.

The CI here also fails with error 429. It means that `https://eth.llamarpc.com` is not the issue.

Let's merge this first: https://github.com/risechain/pevm/pull/333

kien-rise: The CI in #333 also fails. It means that both #332 and #333 must be used to resolve the 429 errors.

I am merging the two PRs.

hai-rise: This is great! Can you merge `self.runtime.block_on(with_retires(...))` into `RpcStorage::fetch` right below `new`? Then apply it to `block_hash` as well.

kien-rise: > This is great! Can you merge `self.runtime.block_on(with_retires(...))` into `RpcStorage::fetch` right below `new`? Then apply it to `block_hash` as well.

You can have a look at the latest commit.

Personally, I recommend reverting this change because the code looks uglier. Note that the output of `self.provider.get_block_by_number` is not `Clone`.

kien-rise: Based on my observations, the frequency of calls to `block_hash` is very rare. Therefore, it unlikely causes error 429.

hai-rise: No, consistency is important. `block_hash` is an unlimited op-code and is frequently used for on-chain randomness. There's no reason to arbitrarily treat it as second-class.

Ugliness would only be a problem if we could write a function without any type parameter. Going from 2 to 4 while eliminating the `Clone` constraint with virtually no extra code lines is fine. Having random inconsistent logic is not.

kien-rise: ok, you can merge this PR if everything looks good

hai-rise: Ok. If readability is important let me add the missing `Clone` upstream. Personally, I find this pretty normal in Rust, and nowhere near Haskell's Monad Transformers and friends.

hai-rise (approved):

PR #333, opened 2024-09-05
feat: add with_retries in src/storage/rpc.rs

kien-rise:
Resolves https://github.com/risechain/pevm/pull/233#discussion_r1745013758

Avoid `429 Too Many Requests` by allowing retries when making RPC requests.


kien-rise:
Included in https://github.com/risechain/pevm/pull/332

PR #337, opened 2024-09-05
feat(seq): use `State` over `CacheDB`

hai-rise:


hai-rise:
`State` is just too heavyweight for our use case.

PR #346, opened 2024-09-09
feat: ERC20 lazy-update

kien-rise:


hai-rise:
This is still a very important idea, but maybe let's reconsider this after having an inline EVM implementation, to keep it low-level, hence lower friction.

PR #347, opened 2024-09-10
fix: use spec_id instead of block.header.number to check if spec Byzantium is enabled

kien-rise:


tests/common/runner.rs
```diff
@@ -66,7 +66,7 @@ pub fn test_execute_alloy<S: Storage + Send + Sync, C: PevmChain + Send + Sync +
         // Before EIP-658 (https://eips.ethereum.org/EIPS/eip-658), the
         // receipt root is calculated with the post transaction state root,
         // which we don't have in these tests.
-        if block.header.number >= 4370000 {
+        if spec_id >= SpecId::BYZANTIUM {
```
hai-rise, line 69: This temporary fix will only last to the OP PR as we parameterize this function by `Chain` but most chains don't have the concept of `BYZANTIUM`. The better fix (for now) is to make `calculate_receipt_root` return an `Option` and only make the assertion if it returns a `Some` here. We then move this EIP-658 logic to Ethereum's implementation of `calculate_receipt_root` and return a `None` (currently unsupported) pre-BYZANTIUM. New chain users can also return a `None` until/unless they need a receipt test/integration.

kien-rise: I changed the function signature to:

```rust
    /// Calculate receipt root
    fn calculate_receipt_root(
        &self,
        spec_id: SpecId,
    ) -> Result<Option<B256>, Self::CalculateReceiptRootError>;
```

Can you help to review again? thanks!

src/chain.rs
```diff
@@ -69,7 +72,7 @@ pub trait PevmChain: Debug {
         spec_id: SpecId,
         txs: &BlockTransactions<Self::Transaction>,
         tx_results: &[PevmTxExecutionResult],
-    ) -> B256;
+    ) -> Result<Option<B256>, Self::CalculateReceiptRootError>;
```
hai-rise, line 75: This is more complicated than needed. If we promote it to `Result` then the `None` should be an `Error`, like `MissingPrevStateRoot`, `Unsupported`, or even `()`.

src/chain/ethereum.rs
```diff
@@ -57,6 +57,12 @@ pub enum EthereumTransactionParsingError {
     MissingMaxFeePerGas,
 }
 
+#[derive(Debug, Clone, PartialEq)]
+pub enum EthereumCalculateReceiptRootError {
+    /// [tx.type] is invalid.
+    InvalidType(u8),
```
hai-rise, line 63: `EthereumTransactionParsingError::InvalidType` is self-described, but here it should be `InvalidTxType`.

src/chain/ethereum.rs
```diff
@@ -189,28 +196,43 @@ impl PevmChain for PevmEthereum {
...
-        let tx_type_iter = txs
-            .txns()
-            .map(|tx| TxType::try_from(tx.transaction_type.unwrap_or_default()).unwrap());
+    ) -> Result<Option<B256>, Self::CalculateReceiptRootError> {
+        if spec_id < SpecId::BYZANTIUM {
+            // We can only calculate the receipts root from Byzantium.
+            // Before EIP-658 (https://eips.ethereum.org/EIPS/eip-658), the
+            // receipt root is calculated with the post transaction state root,
+            // which we don't have in these tests.
+            return Ok(None);
```
hai-rise, line 208: This should be an error, a proper `Ok` would require an extra data type here. Can also promote the comment to a `TODO` (now that we're actively looking for community contributions).

src/chain.rs
kien-rise: ok, let me propose another version that use `Result<B256, CalculateReceiptRootError>`

tests/common/runner.rs
```diff
@@ -62,15 +62,10 @@ pub fn test_execute_alloy<S: Storage + Send + Sync, C: PevmChain + Send + Sync +
...
-        // which we don't have in these tests.
-        if block.header.number >= 4370000 {
-            assert_eq!(
-                block.header.receipts_root,
-                chain.calculate_receipt_root(spec_id, &block.transactions, &tx_results)
-            );
+        match chain.calculate_receipt_root(spec_id, &block.transactions, &tx_results) {
+            Ok(None) => {}
+            Ok(Some(receipt_root)) => assert_eq!(block.header.receipts_root, receipt_root),
+            Err(err) => panic!("{:?}", err),
```
hai-rise, line 68: This is stricter than needed since we cannot guarantee to get a receipt root out of all tests. We can just do this as "if we can get a receipt root here then assert it". Otherwise, this case (receipt root calculation correctness) should be included in dedicated unit tests of each chain.

kien-rise: Can you check my last commit? Is this solution ok?

hai-rise: Not perfect but definitely good enough for now, let's merge.

kien-rise: wait, let me resolve all the discussions first. Some are not reflected in the code.

kien-rise: > Not perfect

Yeah, I understand that this solution does not allow other chain to customize the `Custom` arm. Using `String` is not perfect.

src/chain/ethereum.rs
kien-rise: can you check my latest commit? here is the modified part:

```rust
        if spec_id < SpecId::BYZANTIUM {
            // We can only calculate the receipts root from Byzantium.
            // Before EIP-658 (https://eips.ethereum.org/EIPS/eip-658), the
            // receipt root is calculated with the post transaction state root,
            // which we don't have in these tests.

            // TODO: allow to calculate the receipt root by providing the post transaction state root
            return Err(CalculateReceiptRootError::Unsupported);
        }
```

src/chain/ethereum.rs
```diff
@@ -189,28 +189,44 @@ impl PevmChain for PevmEthereum {
...
-        // 1. Create an iterator of ReceiptEnvelope
-        let tx_type_iter = txs
-            .txns()
-            .map(|tx| TxType::try_from(tx.transaction_type.unwrap_or_default()).unwrap());
+    ) -> Result<B256, CalculateReceiptRootError> {
+        if spec_id < SpecId::BYZANTIUM {
+            // We can only calculate the receipts root from Byzantium.
+            // Before EIP-658 (https://eips.ethereum.org/EIPS/eip-658), the
+            // receipt root is calculated with the post transaction state root,
+            // which we don't have in these tests.
```
hai-rise, line 200: `in these tests` -> `here`.

PR #349, opened 2024-09-11
feat: check EIPs in PevmChain

kien-rise:
We move the EIP check to PevmChain so that we can customize other chains.

### DIFF
--- a/src/chain.rs
+++ b/src/chain.rs
@@ -27,7 +27,7 @@ pub enum CalculateReceiptRootError {
     Unsupported,
     /// Invalid transaction type
     InvalidTxType(u8),
-    /// An escape hatch for other chains
+    /// Arbitrary error message
     Custom(String),
 }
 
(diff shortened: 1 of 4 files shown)


src/chain.rs
```diff
@@ -81,6 +81,18 @@ pub trait PevmChain: Debug {
         txs: &BlockTransactions<Self::Transaction>,
         tx_results: &[PevmTxExecutionResult],
     ) -> Result<B256, CalculateReceiptRootError>;
+
+    /// Check whether EIP-1559 is enabled
+    /// https://github.com/ethereum/EIPs/blob/96523ef4d76ca440f73f0403ddb5c9cb3b24dcae/EIPS/eip-1559.md
+    fn is_eip_1559_enabled(&self, spec_id: SpecId) -> bool {
+        spec_id >= SpecId::LONDON
```
hai-rise, line 88: This must be the Ethereum implementation. Again, most other chains don't have the concept of `LONDON`. This is like using Ethereum's hardfork rules to implement the default `get_block_spec`. Different chains may even need to check their `id` and extra information provided by the `struct` to answer. 

This should either have no default implementation or be a `true`, with the assumption that new modern EVM chains would adopt (these) old EIPs by default and must re-implement otherwise.

kien-rise: removed the default impl

Issue #318, opened 2024-09-04
Rewrite `VmExecutionResult` as `Result`


hai-rise:
https://github.com/risechain/pevm/blob/24266be8d317bdcbe8f8ac10c6ce34ae7298802d/src/vm.rs#L69-L81

hai-rise:
Done in #354.

PR #354, opened 2024-09-15
chore(`vm`): rewrite `VmExecutionResult` as `Result`

EdwardJES:
This PR addresses https://github.com/risechain/pevm/issues/318

### DIFF
--- a/src/vm.rs
+++ b/src/vm.rs
@@ -71,17 +71,6 @@ impl PevmTxExecutionResult {
     }
 }
 
-// /// The error type of [PevmChain::calculate_receipt_root]
-// #[derive(Debug, Clone)]
-// pub enum CalculateReceiptRootError {
-//     /// Unsupported
-//     Unsupported,
-//     /// Invalid transaction type
-//     InvalidTxType(u8),
-//     /// Arbitrary error message
-//     Custom(String),
-// }
-
 pub(crate) enum VmExecutionError {
     Retry,
     FallbackToSequential,
--- a/src/vm.rs
+++ b/src/vm.rs
@@ -279,7 +279,7 @@ impl<'a, S: Storage, C: PevmChain> Database for VmDb<'a, S, C> {
                 loop {
                     match iter.next_back() {
                         Some((blocking_idx, MemoryEntry::Estimate)) => {
-                            return Err(ReadError::BlockingIndex(*blocking_idx))
+                            return Err(ReadError::Blocking(*blocking_idx))
                         }
                         Some((closest_idx, MemoryEntry::Data(tx_incarnation, value))) => {
                             // About to push a new origin
@@ -377,7 +377,7 @@ impl<'a, S: Storage, C: PevmChain> Database for VmDb<'a, S, C> {
                 if self.tx_idx > 0 {
                     // TODO: Better retry strategy -- immediately, to the
                     // closest sender tx, to the missing sender tx, etc.
-                    return Err(ReadError::BlockingIndex(self.tx_idx - 1));
+                    return Err(ReadError::Blocking(self.tx_idx - 1));
                 } else {
                     return Err(ReadError::InvalidNonce(self.tx_idx));
                 }
@@ -462,9 +462,7 @@ impl<'a, S: Storage, C: PevmChain> Database for VmDb<'a, S, C> {
                             )?;
                             return Ok(*value);
                         }
-                        MemoryEntry::Estimate => {
-                            return Err(ReadError::BlockingIndex(*closest_idx))
-                        }
+                        MemoryEntry::Estimate => return Err(ReadError::Blocking(*closest_idx)),
                         _ => return Err(ReadError::InvalidMemoryValueType),
                     }
                 }
--- a/src/vm.rs
+++ b/src/vm.rs
@@ -520,8 +520,9 @@ impl<'a, S: Storage, C: PevmChain> Vm<'a, S, C> {
         let mut db = match VmDb::new(self, tx_version.tx_idx, tx, from_hash, to_hash) {
             Ok(db) => db,
             Err(err) => {
-                let evm_err = ExecutionError::Database(err);
-                return VmExecutionResult::Err(VmExecutionError::ExecutionError(evm_err));
+                return VmExecutionResult::Err(VmExecutionError::ExecutionError(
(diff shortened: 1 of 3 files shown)


src/vm.rs
```diff
@@ -518,8 +519,11 @@ impl<'a, S: Storage, C: PevmChain> Vm<'a, S, C> {
         // Execute
         let mut db = match VmDb::new(self, tx_version.tx_idx, tx, from_hash, to_hash) {
             Ok(db) => db,
-            // TODO: Handle different errors differently
-            Err(_) => return VmExecutionResult::FallbackToSequential,
+            Err(err) => {
+                return VmExecutionResult::Err(VmExecutionError::ExecutionError(
```
EdwardJES, line 523: Hey @hai-rise, I'm not the biggest fan of the stuttering and verbosity here. 

An alternative could impl method e.g.

```
impl VmExecutionError {
    pub fn execution_error(error: ExecutionError) -> Self {
        VmExecutionError::ExecutionError(error)
    }
}
```
Please lmk what you think 🙏

hai-rise: I agree! I think a clean solution is to:
- Bring `ReadError` from `lib.rs` here as a `pub(crate)` (as a related chore).
- Implement a `From<ReadError>`  for `VmExecutionError` as the source-of-truth for read error handling behaviour.
```rust
impl From<ReadError> for VmExecutionError {
    fn from(err: ReadError) -> Self {
        match err {
            ReadError::InconsistentRead => VmExecutionError::Retry,
            ReadError::SelfDestructedAccount => VmExecutionError::FallbackToSequential,
            ReadError::Blocking(tx_idx) => VmExecutionError::Blocking(tx_idx),
            _ => VmExecutionError::ExecutionError(EVMError::Database(err)),
        }
    }
}
```
- Then we can simplify this instance to:
```rust
let mut db = VmDb::new(self, tx_version.tx_idx, tx, from_hash, to_hash)
            .map_err(VmExecutionError::from)?;
```
- And the several match arms below to:
```rust
Err(EVMError::Database(read_error)) => Err(read_error.into()),
```

Also, note that the current change is a semantic change (from returning `FallbackToSequential` to returning `ExecutionError`) that fails test.

src/vm.rs
```diff
@@ -70,20 +71,20 @@ impl PevmTxExecutionResult {
...
-    },
+    ReadError { blocking_tx_idx: TxIdx },
     ExecutionError(ExecutionError),
-    Ok {
-        execution_result: PevmTxExecutionResult,
-        flags: FinishExecFlags,
-    },
 }
 
+pub(crate) struct VmExecutionOutcome {
```
hai-rise, line 81: Naming here is hard! We can keep `VmExecutionResult` for now.

src/vm.rs
```diff
@@ -70,20 +71,20 @@ impl PevmTxExecutionResult {
...
+    pub execution_result: PevmTxExecutionResult,
```
hai-rise, line 82: Let's do `pub(crate)` here.

src/vm.rs
```diff
@@ -70,20 +71,20 @@ impl PevmTxExecutionResult {
...
+    pub flags: FinishExecFlags,
+}
+
+pub(crate) type VmExecutionResult = Result<VmExecutionOutcome, VmExecutionError>;
```
hai-rise, line 86: Inline `Result<VmExecutionResult, VmExecutionError>` to the signature of `Vm::execute` is cleaner.

src/vm.rs
```diff
@@ -646,21 +650,25 @@ impl<'a, S: Storage, C: PevmChain> Vm<'a, S, C> {
                     flags |= FinishExecFlags::WroteNewLocation;
                 }
 
-                VmExecutionResult::Ok {
+                let outcome = VmExecutionOutcome {
```
hai-rise, line 653: Just wrap this in an `Ok`.

hai-rise (changes requested):
I've left a few suggestions. Thank you again for the work!

Also, it'd be nice if this PR also adds `Result<(), VmExecutionError>` to `apply_rewards`, to remove the `unreachable!` panic to: 
```rust
_ => return Err(ReadError::InvalidMemoryValueType.into()),
```

🙏

EdwardJES:
Thank you for the quick feedback @hai-rise, this is ready for another review 🙏

src/vm.rs
```diff
@@ -70,18 +71,51 @@ impl PevmTxExecutionResult {
...
+    /// TODO: Handle this at the type level?
+    InvalidMemoryValueType,
+}
+
+impl From<ReadError> for VmExecutionError {
+    fn from(err: ReadError) -> Self {
+        match err {
+            ReadError::InconsistentRead => VmExecutionError::Retry,
+            ReadError::SelfDestructedAccount => VmExecutionError::FallbackToSequential,
+            ReadError::BlockingIndex(tx_idx) => VmExecutionError::Blocking {
```
hai-rise, line 108: Very minor but we can unify both as `Blocking(TxIdx)` for consistency now 🙏.

hai-rise (changes requested):
Just one minor change left, otherwise it's perfect. Thank you!

EdwardJES:
Thank you @hai-rise, unified the change. I'm looking forward to making more contributions.

EdwardJES:
Thank you @hai-rise, unified the change. I'm looking forward to making more contributions.

src/vm.rs
```diff
@@ -70,18 +71,51 @@ impl PevmTxExecutionResult {
...
 
-// TODO: Rewrite as [Result]
-pub(crate) enum VmExecutionResult {
+pub(crate) enum VmExecutionError {
     Retry,
     FallbackToSequential,
-    ReadError {
-        blocking_tx_idx: TxIdx,
-    },
+    Blocking { blocking_tx_idx: TxIdx },
```
hai-rise, line 77: Sorry, I should have been clearer. But we should also change this to `Blocking(TxIdx)`, so no more `blocking_tx_idx` for both error types.

hai-rise (changes requested):

EdwardJES:
No worries, @hai-rise. I probably should've caught that list nit.

hai-rise (approved):
Better structure and error handling, thank you!

Issue #310, opened 2024-09-04
Post-processing lazy evaluation: assert that the evaluated nonce matches the transaction's


hai-rise:
https://github.com/risechain/pevm/blob/1fe66657fb5ec96663745d60cfd9da2ac18e304a/src/pevm.rs#L259-L296

We can `debug_assert_eq!` for now, to have a round of converting them to error types later.

Issue #341, opened 2024-09-06
Add a `maxperf` mode/flag that skip less serious checks


hai-rise:
Like EIP-7610.

Issue #8, opened 2024-04-19
Robust error handling


hai-rise:
Libraries should not casually panic! We should always return errors to library users and perhaps add an option to fall back to sequential if the parallel executor panics.
Let's improve error types and hunt down them `unwrap`, `unreacheable`, `debug_assert!`, etc.

PR #355, opened 2024-09-15
chore(`pevm`): Post-processing lazy evaluation: assert that the evaluated nonce matches the transaction's

EdwardJES:
This PR addresses https://github.com/risechain/pevm/issues/310

### DIFF
--- a/src/pevm.rs
+++ b/src/pevm.rs
@@ -253,6 +253,7 @@ impl Pevm {
                 };
 
                 for (tx_idx, memory_entry) in write_history.iter() {
+                    let tx = unsafe { txs.get_unchecked(*tx_idx) };
                     match memory_entry {
                         MemoryEntry::Data(_, MemoryValue::Basic(info)) => {
                             // We fall back to sequential execution when reading a self-destructed account,
@@ -271,7 +272,6 @@ impl Pevm {
                             // TODO: Guard against overflows & underflows
                             // Ideally we would share these calculations with revm
                             // (using their utility functions).
-                            let tx = unsafe { txs.get_unchecked(*tx_idx) };
                             let mut max_fee = U256::from(tx.gas_limit) * tx.gas_price + tx.value;
                             if let Some(blob_fee) = tx.max_fee_per_blob_gas {
                                 max_fee +=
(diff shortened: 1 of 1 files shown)


EdwardJES:
Hey @hai-rise, I've tried to have a crack at this one. And I'm not sure if I've identified a bug or not (thinking possibly yes as the other `MemoryValue::Basic` checks pass).

The `mainnet_blocks_from_disk` is failing. I'd love to try and debug this more to better understand the stack, but I may need some pointers if that's okay 🙏 .

Just for my understanding of what's going on here: this is essentially validating the incarnation (tx) that resulted in the particular state mutation of an account, which is encoded in the `write_history`.  Where the `write_history` holds all particular state updates for a given memory location from the resulting incarnations?

Here is the particular `tx_env` where this check fails:

```
TX TxEnv {
    caller: 0xa9cefd4474086c52f15abc0e604284440768da57,
    gas_limit: 23100,
    gas_price: 54641531805,
    transact_to: Call(
        0xcff0b829fd9bd7a58ca8d62eef8105506f859592,
    ),
    value: 500000000000000,
    data: 0x,
    nonce: Some(
        18,
    ),
    chain_id: Some(
        1,
    ),
    access_list: [],
    gas_priority_fee: Some(
        3358468195,
    ),
    blob_hashes: [],
    max_fee_per_blob_gas: None,
    authorization_list: None,
}
```

Tyvm.

src/pevm.rs
```diff
@@ -252,13 +252,23 @@ impl Pevm {
                     None
                 };
 
-                // TODO: Assert that the evaluated nonce matches the tx's.
                 for (tx_idx, memory_entry) in write_history.iter() {
                     match memory_entry {
                         MemoryEntry::Data(_, MemoryValue::Basic(info)) => {
+                            // TODO: these debug assertions can later be converted to errors
```
hai-rise, line 258: We have two issues tracking it!
- https://github.com/risechain/pevm/issues/8
- https://github.com/risechain/pevm/issues/341

We'll have a round of converting these panics into proper error types. Another round to disable less serious checks for people who "own" the input data and want the most competitive speed.

So I think we don't need to add this comment here, otherwise for completeness we should add to all others too!

hai-rise (changes requested):
Ah, `tx.nonce` only applies to the sender, while we can have the recipient (or more addresses in the future) in `MemoryValue::Basic` and other arms.

I think we can add this line below the `match memory_entry`:
```rust
assert!(tx.caller != address || tx.nonce.map_or(true, |n| n + 1 == nonce));
```

EdwardJES:
Thank you for the feedback @hai-rise. That makes sense. Was way off the mark with that bug... Thank you for the learning here.

hai-rise (approved):
Thank you!

Issue #249, opened 2024-08-07
Write a CLI that snapshots an input block number (via RPC storage) to disk


hai-rise:

hai-rise:
Done in #357.

PR #357, opened 2024-09-18
 feat(`cli`): move snapshot functionality to a cli

EdwardJES:
This PR addresses https://github.com/risechain/pevm/issues/249.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,9 +48,9 @@ reqwest = "0.12.7"
 tokio = { version = "1.40.0", features = ["rt-multi-thread"] }
 
 # CLI dependencies
-serde_json = { version = "1.0.127", features = ["preserve_order"] }
 bincode = "1.3.3"
-clap = { version = "4.5.17", features = ["derive"] }
+clap = { version = "4.5.18", features = ["derive"] }
+serde_json = { version = "1.0.128", features = ["preserve_order"] }
 
 # OP dependencies
 op-alloy-consensus = { version = "0.2.12", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,10 @@ optimism = [
     "dep:op-alloy-rpc-types",
 ]
 
+[[bin]]
+name = "fetch"
+path = "bin/fetch.rs"
+
 [dependencies]
 # Put this behind a feature flag if there are use cases & users
 # that need more security guarantees even for internally hashing
@@ -43,20 +47,23 @@ alloy-transport-http = "0.3.6"
 reqwest = "0.12.7"
 tokio = { version = "1.40.0", features = ["rt-multi-thread"] }
 
+# CLI dependencies
+serde_json = { version = "1.0.127", features = ["preserve_order"] }
+bincode = "1.3.3"
+clap = { version = "4.5.17", features = ["derive"] }
+
 # OP dependencies
 op-alloy-consensus = { version = "0.2.12", optional = true }
 op-alloy-network = { version = "0.2.12", optional = true }
 op-alloy-rpc-types = { version = "0.2.12", optional = true }
 
 [dev-dependencies]
-bincode = "1.3.3"
 criterion = "0.5.1"
 rand = "0.8.5"
 rayon = "1.10.0"
 revme = { git = "https://github.com/risechain/revm", rev = "1ad0b38dd50e421b057d6cd73d2e18de04fad095" }
 rpmalloc = { version = "0.2.2", features = ["thread_cache", "global_cache"] }
 snmalloc-rs = "0.3.6"
-serde_json = "1.0.128"
 tikv-jemallocator = "0.6.0"
 walkdir = "2.5.0"
 
--- a/README.md
+++ b/README.md
@@ -60,9 +60,21 @@ Finally, while Aptos and Polygon embed their pevm implementation directly into t
 - Hyper-optimise at low system levels (kernel configurations, writing hot paths in Assembly, etc.).
 - Propose an EIP to "tax" blocks with low parallelism.
 
-```
+```sh
 $ cargo build
 ```
+
+### Tooling
+
+Fetcher
+
+We provide a command-line interface (CLI) tool to snapshot the state of a block. This tool fetches a block from an RPC provider and snapshots the state to disk.
+
+```sh
+$ cargo run --bin fetch <BLOCK_ID> <RPC_URL>
+```
+
+Where `<BLOCK_ID>` may be a hash or a number.
 
 ## Testing
 
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ $ cargo build
 - Support multiple EVM executors (REVM, JIT & AOT compilers, etc.).
 - Hyper-optimise at low system levels (kernel configurations, writing hot paths in Assembly, etc.).
 - Propose an EIP to "tax" blocks with low parallelism.
-q
+
 ### Tooling
 
 #### Fetch real blocks
--- a/README.md
+++ b/README.md
@@ -31,6 +31,10 @@ Finally, while Aptos and Polygon embed their pevm implementation directly into t
 
 - Install [cmake](https://cmake.org) to build `snmalloc` (a highly performant memory allocator).
 
+```sh
+$ cargo build
+```
+
 ### Alpha Done
 
 - Build a Block-STM foundation to improve on.
@@ -60,15 +64,11 @@ Finally, while Aptos and Polygon embed their pevm implementation directly into t
 - Hyper-optimise at low system levels (kernel configurations, writing hot paths in Assembly, etc.).
 - Propose an EIP to "tax" blocks with low parallelism.
 
-```sh
-$ cargo build
-```
-
 ### Tooling
 
-Fetcher
+#### Fetch real blocks
 
-We provide a command-line interface (CLI) tool to snapshot the state of a block. This tool fetches a block from an RPC provider and snapshots the state to disk.
+We often want to fetch real blocks for testing and benchmarking. The `fetch` CLI snapshots everything needed to execute an Ethereum mainnet block to `data`. More networks will be supported in the future.
 
 ```sh
 $ cargo run --bin fetch <BLOCK_ID> <RPC_URL>
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ $ cargo build
 - Support multiple EVM executors (REVM, JIT & AOT compilers, etc.).
 - Hyper-optimise at low system levels (kernel configurations, writing hot paths in Assembly, etc.).
 - Propose an EIP to "tax" blocks with low parallelism.
-
+q
 ### Tooling
 
 #### Fetch real blocks
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ $ cargo build
 We often want to fetch real blocks for testing and benchmarking. The `fetch` CLI snapshots everything needed to execute an Ethereum mainnet block to `data`. More networks will be supported in the future.
 
 ```sh
-$ cargo run --bin fetch <BLOCK_ID> <RPC_URL>
+$ cargo run --bin fetch <RPC_URL> <BLOCK_ID>
 ```
 
 Where `<BLOCK_ID>` may be a hash or a number.
--- a/bin/fetch.rs
+++ b/bin/fetch.rs
@@ -39,7 +39,7 @@ pub fn main() {
         .expect("Failed to fetch block from provider")
         .unwrap_or_else(|| panic!("No block found for ID: {:?}", block_id));
 
-    // TODO: parameterize chain to later add support for `OP`, `RISE`.
+    // TODO: parameterize chain to later add support for `OP`, `RISE`. This will need to contain the spec as well.
     let chain = PevmEthereum::mainnet();
     let spec_id = chain.get_block_spec(&block.header).unwrap();
     let storage = RpcStorage::new(provider, spec_id, BlockId::number(block.header.number - 1));
--- a/bin/fetch.rs
+++ b/bin/fetch.rs
@@ -79,7 +79,7 @@ pub fn main() {
     let file_state = File::create(format!("{dir}/pre_state.json")).unwrap();
     let json_state = serde_json::to_value(&state).unwrap();
     serde_json::to_writer(file_state, &json_state).unwrap();
-    let file_bytecodes = File::create(format!("data/bytecodes.bincode")).unwrap();
+    let file_bytecodes = File::create("data/bytecodes.bincode").unwrap();
     bincode::serialize_into(file_bytecodes, &bytecodes).unwrap();
 
     // Write block hashes to disk.
--- a/bin/fetch.rs
+++ b/bin/fetch.rs
@@ -33,6 +33,7 @@ fn main() {
         eprintln!("{}", err);
         std::process::exit(2);
     }
+    std::process::exit(0)
 }
 
 // TODO: async main?
@@ -55,36 +56,36 @@ fn try_main() -> Result<()> {
 
     // TODO: parameterize `chain` to add support for `OP`, `RISE`, and more.
     let chain = PevmEthereum::mainnet();
-    let spec_id = chain.get_block_spec(&block.header).unwrap_or_else(|e| {
-        panic!(
+    let spec_id = chain.get_block_spec(&block.header).map_err(|err| {
+        format!(
             "Failed to get block spec for block: {}. {:?}",
-            block.header.number, e
+            block.header.number, err
         )
-    });
+    })?;
     let storage = RpcStorage::new(provider, spec_id, BlockId::number(block.header.number - 1));
 
     // Execute the block and track the pre-state in the RPC storage.
     let _ = Pevm::default()
         .execute(&storage, &chain, block.clone(), NonZeroUsize::MIN, true)
-        .unwrap_or_else(|e| panic!("Failed to execute block: {:?}", e));
+        .map_err(|err| format!("Failed to execute block: {:?}", err));
 
     let block_dir = format!("data/blocks/{}", block.header.number);
 
     // Create block directory.
     fs::create_dir_all(block_dir.clone())
-        .unwrap_or_else(|e| panic!("Failed to create block directory: {e}"));
+        .map_err(|err| format!("Failed to create block directory: {err}"))?;
 
     // Create blockfile.
     let block_file = File::create(format!("{block_dir}/block.json"))
-        .unwrap_or_else(|e| panic!("Failed to create block file: {e}"));
+        .map_err(|err| format!("Failed to create block file: {err}"))?;
     serde_json::to_writer(block_file, &block)
-        .unwrap_or_else(|e| panic!("Failed to write block to file: {e}"));
+        .map_err(|err| format!("Failed to write block to file: {err}"))?;
 
     // Populate bytecodes and state from RPC storage.
     let mut state = BTreeMap::<Address, EvmAccount>::new();
     let mut bytecodes: BTreeMap<B256, EvmCode> = match File::open("data/bytecodes.bincode") {
         Ok(file) => bincode::deserialize_from(BufReader::new(file))
-            .unwrap_or_else(|e| panic!("Failed to deserialize bytecodes from file: {e}")),
+            .map_err(|err| format!("Failed to deserialize bytecodes from file: {err}"))?,
         Err(_) => BTreeMap::new(),
     };
     bytecodes.extend(storage.get_cache_bytecodes());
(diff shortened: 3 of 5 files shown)


bin/cli.rs
```diff
@@ -0,0 +1,126 @@
...
+use clap::{Parser, Subcommand};
+use pevm::{
+    chain::{PevmChain, PevmEthereum},
+    EvmAccount, EvmCode, Pevm, RpcStorage,
+};
+use reqwest::Url;
+use tokio::runtime;
+
+#[derive(Parser, Debug)]
+pub struct Cli {
```
EdwardJES, line 23: @hai-rise, naming is indeed hard here. I was thinking that this CLI could be extended in the future to provide further functionality, e.g. could spin up a local anvil instance to run in CI.

Would appreciate any thoughts on on naming here. Thank you.

bin/cli.rs
```diff
@@ -0,0 +1,126 @@
...
+            );
+
+            // Retrieve block to snapshot.
+            let runtime = runtime::Runtime::new().unwrap();
+            let block = runtime
+                .block_on(provider.get_block(block_id, BlockTransactionsKind::Full))
+                .unwrap_or_else(|e| panic!("Failed to fetch block from provider: {e}",))
+                .unwrap_or_else(|| panic!("No block found for ID: {:?}", block_id));
+
+            // TODO: We could support more specs here. This could be another CLI argument.
```
EdwardJES, line 66: I'm not too sure if it matters as I've only seen this in tests, but chainspec assumes that all hardforks are enabled, which is not a case for all EVM chains. Meaning that the pre-compiles pulled in based on the spec may not exist yet. This could lead to undefined behaviour at the VM level.

bin/cli.rs
```diff
@@ -0,0 +1,126 @@
...
+            for (address, mut account) in storage.get_cache_accounts() {
+                if let Some(code) = account.code.take() {
+                    assert_ne!(account.code_hash.unwrap(), KECCAK_EMPTY);
+                    bytecodes.insert(account.code_hash.unwrap(), code);
+                }
+                state.insert(address, account);
+            }
+
+            let file_state = File::create(format!("{dir}/pre_state.json")).unwrap();
+            // Convert BtreeMap to serde IndexMap to provide consistent ordering.
```
EdwardJES, line 111: I've added this to attempt to enforce the ordering for ease of diffs. It as behaving deterministically on my machine, but may need someone else to pull this down to verify.

EdwardJES:
Please also keep me honest on the new directory introduced with the cli. Was unsure of the best approach as this is introducing the first binary into the project.

tests/mainnet.rs
```diff
@@ -57,43 +51,21 @@ fn mainnet_blocks_from_rpc() {
         common::test_execute_alloy(&rpc_storage, &chain, block.clone(), true);
 
         // Snapshot blocks (for benchmark)
-        // TODO: Port to a dedicated CLI instead?
         // TODO: Binary formats to save disk?
         if std::env::var("SNAPSHOT_BLOCKS") == Ok("1".to_string()) {
```
hai-rise, line 55: Ah, we can remove this whole logic now. Any usage should go through the CLI :).

data/blocks/1150000/pre_state.json
```diff
@@ -1 +1 @@
...:"0x0","nonce":0},"0x2a65aca4d5fc5b5c859090a6c34d164135398226":{"storage":{},"balance":"0xc0048cbbf4f2bcb1cf","nonce":295302},"0x3b0b657e5dcb40323ddad8b8af9ebc830aa485f8":{"storage":{},"balance":"0x46a3500832bd0000","nonce":0},"0x628c57a073c3f3c7fc300cbea093d6fd9ee0f607":{"storage":{},"balance":"0x1bafa9ee16e78000","nonce":0},"0x8e23ee67d1332ad560396262c48ffbb01f93d052":{"storage":{},"balance":"0x4b74f0837c164e90000","nonce":12},"0x9af09991ad63814e53ffc1bccf213ee74027608b":{"code_hash":"0xe66df5a409bb62d627a95c0df524c26fb53113311c3713a8886634712f5bf8a0","storage":{"0x3":"0x1158c3c9a70e85d8358972810ed984c8e6ffcf0f","0x405787fa12a823e0f2b7631cc41b3ba8828b3321ca811111fa75cd3aa3bb5ad0":"0x1f93b5e75a04b225816935c3d65f552cac88a61b","0x2":"0x39","0x405787fa12a823e0f2b7631cc41b3ba8828b3321ca811111fa75cd3aa3bb5acf":"0xc400b9d93a23b0be5d41ab337ad605988aef8463"},"balance":"0xf4a6","nonce":0},"0x9e21ef129618e859f081c8be4bcf114c5b2c750d":{"storage":{},"balance":"0x454ab058c9828180","nonce":2},"0xc400b9d93a23b0be5d41ab337ad605988aef8463":{"storage":{},"balance":"0x2aa675ad2e375fc78","nonce":36742},"0xe27b000f885a6f0cf8a31b6348167c131cfd24c4":{"storage":{},"balance":"0x11c37937e08000","nonce":0}}
```
hai-rise, line 1: Let's not check in `data` changes in this PR.

tests/mainnet.rs
```diff
@@ -57,43 +51,21 @@ fn mainnet_blocks_from_rpc() {
...
-            bincode::serialize_into(file_bytecodes, &bytecodes).unwrap();
-
-            // We convert to [BTreeMap] for consistent ordering & diffs between snapshots
-            let block_hashes: BTreeMap<u64, B256> =
-                rpc_storage.get_cache_block_hashes().into_iter().collect();
-            if !block_hashes.is_empty() {
-                let file = File::create(format!("{dir}/block_hashes.json")).unwrap();
-                serde_json::to_writer(file, &block_hashes).unwrap();
-            }
+            process::Command::new("cargo")
```
hai-rise, line 56: Let's move this to `README.md` under `Development` instead.

Cargo.toml
```diff
@@ -47,16 +53,15 @@ tokio = { version = "1.40.0", features = ["rt-multi-thread"] }
 op-alloy-consensus = { version = "0.2.7", optional = true }
 op-alloy-network = { version = "0.2.7", optional = true }
 op-alloy-rpc-types = { version = "0.2.7", optional = true }
+clap = { version = "4.5.17", features = ["derive"] }
```
hai-rise, line 56: Can you group CLI dependencies like `RPC Storage dependencies`? In the future, we should make them `optional`, the CLI a package or use Cargo workspace to make the core library as lean as possible.

bin/cli.rs
hai-rise: `preserve_order` only works if we snapshot with sequential execution. Parallel state access can be non-deterministic for complicated blocks. 

Let's keep this as we should do sequential execution for stability anyway. The bottleneck is networking so trading stability for parallel performance isn't worth it.

bin/cli.rs
hai-rise: Yeah, I think a `pevm` CLI would be sick. For now, though, let's keep it simple by naming it something like `fetch` without any sub-command.

bin/cli.rs
```diff
@@ -0,0 +1,126 @@
...
+
+#[derive(Subcommand, Debug)]
+pub enum Commands {
+    /// Snapshot a block to disk.
+    Snapshot {
+        /// The ID of the block.
+        block_id: BlockId,
+        /// The RPC url to connect to.
+        rpc_url: String,
+        /// The output directory to write the snapshot to.
```
hai-rise, line 36: Let's keep it simple by hardcoding `output_dir` and `bytecode_path` to `data/` for now.

bin/cli.rs
```diff
@@ -0,0 +1,126 @@
...
+        output_dir: String,
+        /// Path to bytecodes.
+        bytecode_path: Option<String>,
+    },
+}
+
+pub fn main() {
```
hai-rise, line 43: It'd be nice if we could write this `async` with Tokio, or let's leave a TODO for it.

bin/cli.rs
```diff
@@ -0,0 +1,126 @@
...
+        Commands::Snapshot {
+            block_id,
+            rpc_url,
+            output_dir,
+            bytecode_path,
+        } => {
+            // Define provider.
+            let provider = ProviderBuilder::new().on_http(
+                Url::parse(&rpc_url)
+                    .unwrap_or_else(|_| panic!("Invalid RPC URL supplied: {rpc_url}")),
```
hai-rise, line 56: We can shorten these `.unwrap_or_else` with `.expect`?

bin/cli.rs
```diff
@@ -0,0 +1,126 @@
...
+
+            // Retrieve block to snapshot.
+            let runtime = runtime::Runtime::new().unwrap();
+            let block = runtime
+                .block_on(provider.get_block(block_id, BlockTransactionsKind::Full))
+                .unwrap_or_else(|e| panic!("Failed to fetch block from provider: {e}",))
+                .unwrap_or_else(|| panic!("No block found for ID: {:?}", block_id));
+
+            // TODO: We could support more specs here. This could be another CLI argument.
+            // TODO: There is a a weak link between spec_id, provider and thus RPCStorage.
```
hai-rise, line 67: Perhaps the better `TODO` is to parameterize `chain` that gets passed in with Ethereum Mainnet being the default.

bin/cli.rs
```diff
@@ -0,0 +1,126 @@
...
+            // Not all EVM chains support all hardforks, especially rollups.
+            let chain = PevmEthereum::mainnet();
+            let spec_id = chain.get_block_spec(&block.header).unwrap();
+            let storage =
+                RpcStorage::new(provider, spec_id, BlockId::number(block.header.number - 1));
+
+            // Populate RPC storage.
+            let concurrency_level = thread::available_parallelism().unwrap_or(NonZeroUsize::MIN);
```
hai-rise, line 75: Let's execute sequentially here for stability! The bottleneck is networking so trading stability for parallel's performance isn't worth it.

bin/cli.rs
```diff
@@ -0,0 +1,126 @@
...
+            let mut pevm = Pevm::default();
+            let _ = pevm
+                .execute(&storage, &chain, block.clone(), concurrency_level, false)
+                .unwrap();
+
+            // Create dir.
```
hai-rise, line 81: Let's use the hardcoded `data` structure for simplicity for now.

bin/cli.rs
```diff
@@ -0,0 +1,126 @@
...
+            }
+
+            let file_state = File::create(format!("{dir}/pre_state.json")).unwrap();
+            // Convert BtreeMap to serde IndexMap to provide consistent ordering.
+            let json_state = serde_json::to_value(&state).unwrap();
+            serde_json::to_writer(file_state, &json_state).unwrap();
+            let file_bytecodes = File::create(format!("{output_dir}/bytecodes.bincode")).unwrap();
+            bincode::serialize_into(file_bytecodes, &bytecodes).unwrap();
+
+            // We convert to [BTreeMap] for consistent ordering & diffs between snapshots
```
hai-rise, line 117: Perhaps we don't need this anymore!?

hai-rise (changes requested):
Sorry for the late review, only got back from Token yesterday 🙏.

And thank you for the work, this is very helpful!

bin/cli.rs
hai-rise: Yes, let's leave a TODO to parameterize `chain` to later add support for OP, RISE, etc.

bin/cli.rs
EdwardJES: Yeah, this was my original thought. However clippy doesn't like that format with dynamic arguments:
```
.expect(format!("No block found for ID: {:?}", block_id).as_str());
   |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try: `unwrap_or_else(|| panic!("No block found for ID: {:?}", block_id))`
 ```

File: tests/mainnet.rs (hunk 6 above)
Comment (EdwardJES) [reply tests/mainnet.rs:56]: Sorry, could you elaborate on this part a bit more? Do you want me to add this command in the MD so users know how to run it themselves?

File: bin/cli.rs (hunk 9 above)
Comment (EdwardJES) [reply bin/cli.rs]: Left a todo. The stack is panicking with:

```
Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.
```

I'm assuming this has something to do with the pevm execution stack somewhere. Can look to address in a follow up PR.

File: bin/cli.rs (hunk 10 above)
Comment (hai-rise) [reply bin/cli.rs]: Ah, no worries then, let's just follow `clippy` for consistency 🙏.

File: bin/cli.rs (hunk 9 above)
Comment (hai-rise) [reply bin/cli.rs]: Yeah, we must also make `RpcStorage` functions `async`. Other than that the main pevm execution stack doesn't depend on Tokio!

A future PR is fine, thanks.

File: tests/mainnet.rs (hunk 6 above)
Comment (EdwardJES) [reply tests/mainnet.rs:56]: Added to the readme.md https://github.com/risechain/pevm/pull/357/commits/fdaa244b3664d0473a811dd8190b7dd5e96693ff.

**Comment by EdwardJES** [comment]
Thank you for the comments, @hai-rise. This is ready for another round of reviews. 🙏

**Review by hai-rise — CHANGES_REQUESTED** [review]
Only nit comments left, let's go :pray:!

File: Cargo.toml [hunk 15]
```diff
@@ -43,20 +47,23 @@ alloy-transport-http = "0.3.6"
 reqwest = "0.12.7"
 tokio = { version = "1.40.0", features = ["rt-multi-thread"] }
 
+# CLI dependencies
+serde_json = { version = "1.0.127", features = ["preserve_order"] }
```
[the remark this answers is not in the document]
EdwardJES: Awesome, will bump.

README.md
```diff
@@ -60,10 +60,22 @@ Finally, while Aptos and Polygon embed their pevm implementation directly into t
 - Hyper-optimise at low system levels (kernel configurations, writing hot paths in Assembly, etc.).
 - Propose an EIP to "tax" blocks with low parallelism.
 
-```
+```sh
```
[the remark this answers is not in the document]
EdwardJES: Sometimes this is when the best code is written 😉

README.md
```diff
@@ -60,10 +60,22 @@ Finally, while Aptos and Polygon embed their pevm implementation directly into t
...
 $ cargo build
 ```
 
+### Tooling
```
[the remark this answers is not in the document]
EdwardJES: Yep, I like it.

bin/fetch.rs
```diff
@@ -0,0 +1,91 @@
+#![allow(missing_docs)]
```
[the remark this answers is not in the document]
EdwardJES: Ah yep, my b good catch.

bin/fetch.rs
```diff
@@ -0,0 +1,91 @@
...
+use reqwest::Url;
+use tokio::runtime::Runtime;
+
+#[derive(Parser, Debug)]
+/// Fetch is a CLI tool to fetch a block from an RPC provider, and snapshot that block to disk.
+pub struct Fetch {
+    /// The ID of the block.
+    block_id: BlockId,
+    /// The RPC url to connect to.
+    rpc_url: String,
```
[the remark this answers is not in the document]
EdwardJES: This is a great tip.

bin/fetch.rs
```diff
@@ -0,0 +1,91 @@
...
+    let storage = RpcStorage::new(provider, spec_id, BlockId::number(block.header.number - 1));
+
+    // Populate RPC storage.
+    let concurrency_level = thread::available_parallelism().unwrap_or(NonZeroUsize::MIN);
+    let mut pevm = Pevm::default();
+    let _ = pevm
+        .execute(&storage, &chain, block.clone(), concurrency_level, true)
+        .unwrap();
+
+    let dir = format!("data/blocks/{}", block.header.number);
```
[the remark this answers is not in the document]
EdwardJES: Good catch!

bin/fetch.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    // Create blockfile.
+    let block_file = File::create(format!("{dir}/block.json"))
+        .unwrap_or_else(|e| panic!("Failed to create block file: {e}"));
+    serde_json::to_writer(block_file, &block).expect("Failed to write block to file");
+
+    // Load bytecodes.
```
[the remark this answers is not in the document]
EdwardJES: This is clearer. Agree.

EdwardJES:
Thank you for the comments, @hai-rise. The abundance of panics did not feel idiomatic, I've refactored to use result. Tyvm 🙏

hai-rise (changes requested):
Final changes now :pray:.

I'm using the tool already, thanks :saluting_face:!
So much better than the previous interface that also requires re-compilation per run..

README.md
```diff
@@ -59,10 +63,18 @@ Finally, while Aptos and Polygon embed their pevm implementation directly into t
 - Support multiple EVM executors (REVM, JIT & AOT compilers, etc.).
 - Hyper-optimise at low system levels (kernel configurations, writing hot paths in Assembly, etc.).
 - Propose an EIP to "tax" blocks with low parallelism.
+q
```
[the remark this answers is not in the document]
EdwardJES: Yikes, sorry about that.

bin/fetch.rs
```diff
@@ -0,0 +1,124 @@
...
+use alloy_rpc_types::{BlockId, BlockTransactionsKind};
+use clap::Parser;
+use pevm::{
+    chain::{PevmChain, PevmEthereum},
+    EvmAccount, EvmCode, Pevm, RpcStorage,
+};
+use reqwest::Url;
+use tokio::runtime::Runtime;
+
+type Result<T> = std::result::Result<T, Box<dyn Error>>;
```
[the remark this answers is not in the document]
EdwardJES: Good call. Ty.

bin/fetch.rs
```diff
@@ -0,0 +1,124 @@
...
+    let spec_id = chain.get_block_spec(&block.header).map_err(|err| {
+        format!(
+            "Failed to get block spec for block: {}. {:?}",
+            block.header.number, err
+        )
+    })?;
+    let storage = RpcStorage::new(provider, spec_id, BlockId::number(block.header.number - 1));
+
+    // Execute the block and track the pre-state in the RPC storage.
+    let _ = Pevm::default()
```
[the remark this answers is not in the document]
EdwardJES: Good catch!

bin/fetch.rs
```diff
@@ -0,0 +1,124 @@
...
+fn main() {
+    if let Err(err) = try_main() {
+        eprintln!("{}", err);
+        std::process::exit(2);
+    }
+    std::process::exit(0)
+}
+
+// TODO: async main?
+// TODO: Binary formats to save disk?
```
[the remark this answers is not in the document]
EdwardJES: I can maybe add this in the PR, are you thinking of `debug_assert_eq!` against the block properties?

tests/mainnet.rs
```diff
@@ -55,46 +47,6 @@ fn mainnet_blocks_from_rpc() {
         let spec_id = chain.get_block_spec(&block.header).unwrap();
         let rpc_storage = RpcStorage::new(provider, spec_id, BlockId::number(block_number - 1));
         common::test_execute_alloy(&rpc_storage, &chain, block.clone(), true);
```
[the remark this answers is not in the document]
EdwardJES: Thats a great catch

bin/fetch.rs
```diff
@@ -0,0 +1,124 @@
...
+            assert_ne!(code_hash, KECCAK_EMPTY);
+            bytecodes.insert(code_hash, code);
+        }
+        state.insert(address, account);
+    }
+
+    // Write state and bytecodes to disk.
+    let file_state = File::create(format!("{block_dir}/pre_state.json"))
+        .map_err(|err| format!("Failed to create pre-state file: {err}"))?;
+    let json_state = serde_json::to_value(&state)?;
```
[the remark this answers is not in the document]
EdwardJES: Yeah, best to capture all for ux here. Good point.

bin/fetch.rs
```diff
@@ -0,0 +1,124 @@
...
+    serde_json::to_writer(file_state, &json_state)
+        .map_err(|err| format!("Failed to write pre-state to file: {err}"))?;
+    let file_bytecodes = File::create("data/bytecodes.bincode")
```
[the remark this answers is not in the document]
EdwardJES: Let's go full clean. Much prefer it.

bin/fetch.rs
```diff
@@ -0,0 +1,124 @@
...
+        .execute(&storage, &chain, block.clone(), NonZeroUsize::MIN, true)
+        .map_err(|err| format!("Failed to execute block: {:?}", err));
+
+    let block_dir = format!("data/blocks/{}", block.header.number);
+
+    // Create block directory.
+    fs::create_dir_all(block_dir.clone())
+        .map_err(|err| format!("Failed to create block directory: {err}"))?;
+
+    // Create blockfile.
```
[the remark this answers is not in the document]
EdwardJES: Agree.

EdwardJES:
Great feedback, @hai-rise glad the tool is helping. Excited to help!

bin/fetch.rs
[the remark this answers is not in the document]
hai-rise: Another PR is fine so this one can just focus on "fetching". But yeah, it'd be nice to re-read the snapshotted data to run through `common::test_execute_alloy` to test both the data and parallel execution on it.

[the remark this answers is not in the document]
EdwardJES: Okay, gotcha.

bin/fetch.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    // Write pre-state to disk.
+    let file_state = File::create(format!("{block_dir}/pre_state.json"))
+        .map_err(|err| format!("Failed to create pre-state file: {err}"))?;
+    let json_state = serde_json::to_value(&state)
+        .map_err(|err| format!("Failed to serialize pre-state to JSON: {err}"))?;
+    serde_json::to_writer(file_state, &json_state)
+        .map_err(|err| format!("Failed to write pre-state to file: {err}"))?;
+
+    // Write block to disk.
```
hai-rise, line 108: Final one: `block hashes`!

EdwardJES: I had this and removed it. Duh..

hai-rise (approved):
Let's go!

Issue #317, opened 2024-09-04
Support EIP-7702


hai-rise:
https://github.com/risechain/pevm/blob/24266be8d317bdcbe8f8ac10c6ce34ae7298802d/src/storage.rs#L105

EdwardJES:
Hey @hai-rise, getting Pectra supported is exciting. How do you expect the `EvmCode` struct to evolve here? If this is expected to hold all bytecode formats, I think the best path forward would be adopting a similar API to how revm did it: https://github.com/bluealloy/revm/blob/acf1a70ca8350e5144275769262f59a66700a49d/crates/primitives/src/bytecode.rs#L19. 

We would then have individual structs per bytecode where we can extend the types. 

Did you have a similar pattern in mind?

Ty.

hai-rise:
> Hey @hai-rise, getting Pectra supported is exciting. How do you expect the `EvmCode` struct to evolve here? If this is expected to hold all bytecode formats, I think the best path forward would be adopting a similar API to how revm did it: https://github.com/bluealloy/revm/blob/acf1a70ca8350e5144275769262f59a66700a49d/crates/primitives/src/bytecode.rs#L19.
> 
> We would then have individual structs per bytecode where we can extend the types.
> 
> Did you have a similar pattern in mind?
> 
> Ty.

Yes, and #358 is in the right direction already!

hai-rise:
Done in #358.

Issue #302, opened 2024-09-04
Support Pectra hardfork


hai-rise:
Related: #316, #317.

hai-rise:
Done for now!?

PR #358, opened 2024-09-19
feat(`storage`): add support for eip 7702 bytecode

EdwardJES:
PR addressing https://github.com/risechain/pevm/issues/317, working towards https://github.com/risechain/pevm/issues/302


EdwardJES:
Dam, the new `bytecodes` file has balooned in size due to the new enum field bringing the size to ~ 53mb and GH is throwing warnings 😭 .

src/storage.rs
```diff
@@ -68,26 +68,49 @@ impl Default for AccountBasic {
...
-    /// Original bytes length
+    /// Original bytes length.
     original_len: usize,
     /// Jump table.
     jump_table: Arc<BitVec<u8>>,
 }
 
+#[derive(Debug, Clone, PartialEq, Default, Serialize, Deserialize)]
+pub struct Eip7702Code {
+    /// EOA which which will inherit the bytecode.
```
hai-rise, line 83: Double `which`!

src/storage.rs
```diff
@@ -68,26 +68,49 @@ impl Default for AccountBasic {
     }
 }
 
-/// EVM Code, currently mapping to REVM's [ByteCode::LegacyAnalyzed].
-// TODO: Support raw legacy & EOF
 #[derive(Debug, Clone, PartialEq, Default, Serialize, Deserialize)]
-pub struct EvmCode {
-    /// Bytecode with 32 zero bytes padding
+pub struct LegacyAnalyzedCode {
```
hai-rise, line 72: We can just call this `LegacyCode` that we enforce to be analyzed for performance.

src/storage.rs
```diff
@@ -68,26 +68,49 @@ impl Default for AccountBasic {
...
+    /// Raw bytecode encoding MAGIC, version and delegated address.
+    raw: Bytes,
+}
+
+/// EVM Code, currently mapping to REVM's [ByteCode].
+// TODO: Support raw legacy & EOF
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub enum EvmCode {
+    /// Analyzed legacy code, with jump table.
+    LegacyAnalyzed(LegacyAnalyzedCode),
```
hai-rise, line 96: Just `Legacy` is good as we should map both non & analyzed REVM bytecodes to the same analyzed structure.

src/storage.rs
```diff
@@ -68,26 +68,49 @@ impl Default for AccountBasic {
...
     fn from(code: EvmCode) -> Self {
         // TODO: Better error handling.
         // A common trap would be converting a default [EvmCode] into
         // a [Bytecode]. On failure we should fallback to legacy and
         // analyse again.
-        unsafe {
-            Bytecode::new_analyzed(code.bytecode, code.original_len, JumpTable(code.jump_table))
+        match code {
+            EvmCode::Eip7702(code) => {
+                Bytecode::Eip7702(Eip7702Bytecode::new(code.delegated_address))
```
hai-rise, line 109: Shouldn't we construct a full `Bytecode::Eip7702` by "forwarding" all the fields here? This conversion seems to create an empty code regardless of the current `code`.

data/bytecodes.bincode
hai-rise, line 1: Hmm, I wonder what changes here?

hai-rise (changes requested):
This is very neat, thanks! 

I've left a few comments and adding a test would help a lot 🙏.

src/storage.rs
hai-rise, line 109: I think there are two choices here:
1. For simplicity, make `EvmCode::Eip7702` wrap an `Address` and forward it here. Then perhaps add a `TODO` to support custom versions and magics.
2. Construct the full `Bytecode::Eip7702` by forwarding all the fields here.

Which do you prefer?

src/storage.rs
```diff
@@ -68,26 +68,49 @@ impl Default for AccountBasic {
...
 }
 
+#[derive(Debug, Clone, PartialEq, Default, Serialize, Deserialize)]
+pub struct Eip7702Code {
+    /// EOA which which will inherit the bytecode.
+    delegated_address: Address,
+    /// Version of the bytecode.
+    version: u8,
+    /// Raw bytecode encoding MAGIC, version and delegated address.
+    raw: Bytes,
```
hai-rise, line 88: I don't think we need this for "storage". We can always re-construct it at runtime for execution (like at `revm` level) from `address` and `version`.

data/bytecodes.bincode
EdwardJES: Good question. This is because the bincode encoding has to support the new field I added. I guess this is to be aware when decoding.

hai-rise: Ah, yes, it should add a prefix byte to know which arm/type this bytecode is. Fair, thanks 🙏.

src/storage.rs
EdwardJES: Good callout. I'll go for the latter here.

src/storage.rs
EdwardJES: Good callout. I've removed `raw` and kept `version` so we can preserve it. This has naturally made me choose option 2. from above.

EdwardJES:
Hey @hai-rise, thanks for the feedback 🙏 . I've added tests and addressed the above comments. LMK if you'd like me to add tests converting back to the `revm` types.

tests/bytecode.rs
```diff
@@ -0,0 +1,65 @@
...
+    let evmcode = EvmCode::from(bytecode);
+    assert!(
+        matches!(evmcode, EvmCode::Eip7702(Eip7702Code { delegated_address, version })
+            if delegated_address == addr && version == 0
+        )
+    );
+}
+
+#[test]
+fn test_evmcode_from_revm_bytecode_legacy_raw() {
```
EdwardJES, line 36: I'm in favour of the format: `test_<describe the test>` and it would be great to consistency across the test names. Please lmk what you think.

src/storage.rs
```diff
@@ -68,16 +71,34 @@ impl Default for AccountBasic {
...
+/// Analyzed legacy code.
 #[derive(Debug, Clone, PartialEq, Default, Serialize, Deserialize)]
-pub struct EvmCode {
-    /// Bytecode with 32 zero bytes padding
-    bytecode: Bytes,
-    /// Original bytes length
-    original_len: usize,
+pub struct LegacyCode {
+    /// Bytecode with 32 zero bytes padding.
+    pub bytecode: Bytes,
```
EdwardJES, line 78: `pub(crate)`?

src/storage.rs
```diff
@@ -68,16 +71,34 @@ impl Default for AccountBasic {
     }
 }
 
-/// EVM Code, currently mapping to REVM's [ByteCode::LegacyAnalyzed].
-// TODO: Support raw legacy & EOF
+/// Analyzed legacy code.
 #[derive(Debug, Clone, PartialEq, Default, Serialize, Deserialize)]
```
EdwardJES, line 75: Is `PartialEq` needed here 🤔

hai-rise: Yes, it is required for `EvmCode` -> `EvmAccount` -> `EvmStateTransitions` -> `PevmTxExecutionResult`. It won't be "required" anymore after #202 but comparing bytecodes still sounds convenient. I guess we can also implement more performant `PartialEq`s by comparing only the raw bytecodes, but it's out-of-scope here :pray:.

tests/bytecode.rs
hai-rise: I don't have a strong preference here. Feel the `test_` prefix is a bit redundant as there's usually a `#[test]` above the `fn`, and the log would be:
```
running 3 tests
test test_evmcode_from_revm_bytecode_eip7702 ... ok
test test_evmcode_from_revm_bytecode_legacy_analyzed ... ok
test test_evmcode_from_revm_bytecode_legacy_raw ... ok
```

Instead of just:
```
running 2 tests
test beneficiary_heavy_evaluation ... ok
test beneficiary_random ... ok
```

src/storage.rs
```diff
@@ -68,16 +71,34 @@ impl Default for AccountBasic {
...
-pub struct EvmCode {
-    /// Bytecode with 32 zero bytes padding
-    bytecode: Bytes,
-    /// Original bytes length
-    original_len: usize,
+pub struct LegacyCode {
+    /// Bytecode with 32 zero bytes padding.
```
hai-rise, line 77: The `Eip7702Code`'s raw bytecode conversation made me realize we don't need to store padded bytecodes as well. Perhaps add a `// TODO: Store unpadded bytecode and pad on revm conversion` here?

src/storage.rs
hai-rise: We can remove the `Default`s though, which is also error-prone. Anything that can have an `EvmCode` like `EvmAccount` should use `None` instead of a default code (that is probably invalid anyway).

src/storage.rs
hai-rise, line 78: Let's not make these fields `pub` as that allows people to construct an invalid `LegacyCode` in memory. For now, let's remove the `Default`s and guard the only entry of `revm` conversion well. In the future, we can add `new` functions that do proper validation and analysis and return a `Result`.

src/storage.rs
```diff
@@ -86,23 +107,39 @@ impl From<EvmCode> for Bytecode {
...
+            // This arm will recursively fallback to LegacyAnalyzed.
             Bytecode::LegacyRaw(_) => to_analysed(code).into(),
-            Bytecode::LegacyAnalyzed(code) => EvmCode {
+            Bytecode::LegacyAnalyzed(code) => EvmCode::Legacy(LegacyCode {
                 bytecode: code.bytecode,
                 original_len: code.original_len,
                 jump_table: code.jump_table.0,
-            },
+            }),
             Bytecode::Eof(_) => unimplemented!("TODO: Support EOF"),
```
hai-rise, line 138: Nit but putting `unimplemented!` last is more visible :pray:.

tests/bytecode.rs
```diff
@@ -0,0 +1,73 @@
+use std::str::FromStr;
```
hai-rise, line 1: I think this is cleaner as a 
```
#[cfg(test)]
mod tests {
    use super::*;
}
```

in `storage.rs`. It's more appropriate for unit tests for location, removing the need to make fields `pub`, etc. Most stuff in `tests` are integration tests for the `execute_*` functions.

tests/bytecode.rs
```diff
@@ -0,0 +1,73 @@
...
...5610313576000805b50508051604051600082901a600183901a60081b1760029290921a60101b91909117639e3779b9810260111c617ffc16909101805160e081811c878603811890911b9091189091528401908183039084841061026857506102a3565b600184019350611fff821161029d578251600081901a600182901a60081b1760029190911a60101b17810361029d57506102a3565b5061020c565b8383106102b1575050610313565b600183039250858311156102cf576102cc87878886036100a9565b96505b6102e3600985016003850160038501610079565b91506102f08782846100f7565b9650506103088461030386848601610186565b610186565b915050809350610200565b5050617fe061032884848589518601036100a9565b03925050506020820180820383525b81811161034e57617fe08101518152602001610337565b5060008152602001604052919050565b634e487b7160e01b600052604160045260246000fd5b60006020828403121561038657600080fd5b813567ffffffffffffffff8082111561039e57600080fd5b818401915084601f8301126103b257600080fd5b8135818111156103c4576103c461035e565b604051601f8201601f19908116603f011681019083821181831017156103ec576103ec61035e565b8160405282815287602084870101111561040557600080fd5b82602086016020830137600092810160200192909252509594505050505056fea264697066735822122000646b2953fc4a6f501bd0456ac52203089443937719e16b3190b7979c39511264736f6c63430008190033");
```
hai-rise, line 10: A comment on what this bytecode is would be nice :pray:.

tests/bytecode.rs
```diff
@@ -0,0 +1,73 @@
...
+
+#[test]
+fn test_evmcode_from_revm_bytecode_eip7702() {
```
hai-rise, line 13: Nit but ordering-wise I'd put this last out of 3 functions :sweat_smile:.

tests/bytecode.rs
```diff
@@ -0,0 +1,73 @@
...
+    let addr = Address::new([0x01; 20]);
+
+    // New from address.
+    let bytecode = Bytecode::Eip7702(Eip7702Bytecode::new(addr));
+    let evmcode = EvmCode::from(bytecode);
+    assert!(
+        matches!(evmcode, EvmCode::Eip7702(Eip7702Code { delegated_address, .. })
+            if delegated_address == addr
```
hai-rise, line 21: We can also assert that `version == EIP7702_VERSION` here?

tests/bytecode.rs
```diff
@@ -0,0 +1,73 @@
...
+        )
+    );
+
+    // New fromn raw.
+    let byte_str = format!("ef0100{}", addr.to_string().trim_start_matches("0x"));
```
hai-rise, line 26: We can use `EIP7702_MAGIC_BYTES` and `EIP7702_VERSION` instead of hard-coded values here?

tests/bytecode.rs
```diff
@@ -0,0 +1,73 @@
...
+    let raw = Bytes::from_str(&byte_str).unwrap();
+    let bytecode = Bytecode::Eip7702(Eip7702Bytecode::new_raw(raw).unwrap());
+    let evmcode = EvmCode::from(bytecode);
+    assert!(
+        matches!(evmcode, EvmCode::Eip7702(Eip7702Code { delegated_address, version })
+            if delegated_address == addr && version == 0
```
hai-rise, line 32: `EIP7702_VERSION`?

tests/bytecode.rs
hai-rise: An interesting test we can add is to mutate an `Eip7702Bytecode`'s `version`, convert it to `EvmCode` and assert that the `version` is correctly kept. Would raise an error if a future change hardcodes `version` or something.

tests/bytecode.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    if let Bytecode::LegacyAnalyzed(legacy_analyzed) = analyzed {
+        let raw_jump = legacy_analyzed.jump_table().0.clone();
+        assert!(
+            matches!(evmcode, EvmCode::Legacy(LegacyCode { bytecode, original_len, jump_table })
+                if bytecode == *legacy_analyzed.bytecode() && original_len == legacy_analyzed.original_len() && jump_table == raw_jump
+            )
+        );
+    } else {
+        panic!("Expected LegacyAnalyzed Bytecode")
```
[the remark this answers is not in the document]
hai-rise: We can write something like this to deduplicate code and make these tests much cleaner:
```rust
fn eq_bytecodes(revm_code: &Bytecode, pevm_code: &EvmCode) -> bool {
    match (revm_code, pevm_code) {
        ...
    }
}
```

tests/bytecode.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+}
+
+#[test]
+fn test_evmcode_from_revm_bytecode_legacy_analyzed() {
```
hai-rise, line 57: We can probably merge this with `_raw` to just `_legacy()`.

tests/bytecode.rs
EdwardJES: No problems, will adjust.

src/storage.rs
EdwardJES: Thank you, makes sense.

tests/bytecode.rs
EdwardJES: Good call, not sure what I was thinking here. These are definitely unit tests I'm targetting.

src/storage.rs
EdwardJES: Good callout.

src/storage.rs
```diff
@@ -68,41 +71,73 @@ impl Default for AccountBasic {
...
+    /// Maps both analyzed and non-analyzed REVM legacy bytecode.
+    Legacy(LegacyCode),
+    /// Maps delegated EIP7702 bytecode.
+    Eip7702(Eip7702Code),
+}
+
 impl From<EvmCode> for Bytecode {
     fn from(code: EvmCode) -> Self {
         // TODO: Better error handling.
-        // A common trap would be converting a default [EvmCode] into
```
EdwardJES, line 86: Rm'd this comment as `default` is no longer derived.

EdwardJES:
@hai-rise, thank you for the feedback. Have refactored with the suggestion above 🙏

src/storage.rs
hai-rise: Nice!

hai-rise (changes requested):
A few more nits :pray:.

src/storage.rs
```diff
@@ -193,3 +228,78 @@ mod in_memory;
...
+        match evmcode {
+            EvmCode::Eip7702(Eip7702Code {
+                delegated_address,
+                version: ver,
+            }) => delegated_address == address && *ver == version,
+            _ => false,
+        }
+    }
+
+    // Bytecode from Storage.sol.
```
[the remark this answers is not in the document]
EdwardJES: Yeah apologies, I was thinking that the `storage.sol` reference from the default remix contract would've done the trick. Agree, not clear. I can make this clearer.

src/storage.rs
```diff
@@ -193,3 +228,78 @@ mod in_memory;
...
+                let raw_jump = revm.jump_table().0.clone();
+                revm.bytecode == pevm.bytecode
+                    && revm.original_len == pevm.original_len
+                    && raw_jump == pevm.jump_table
+            }
+            _ => false,
+        }
+    }
+
+    fn eq_eip7702_version_addr(evmcode: &EvmCode, address: &Address, version: u8) -> bool {
```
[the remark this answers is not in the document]
EdwardJES: Good point.

src/storage.rs
```diff
@@ -193,3 +228,78 @@ mod in_memory;
...
+        let mut bytes = EIP7702_MAGIC_BYTES.to_vec();
+        bytes.push(EIP7702_VERSION);
+        bytes.extend(addr);
+        let bytecode = Bytecode::Eip7702(Eip7702Bytecode::new_raw(bytes.clone().into()).unwrap());
+        let evmcode = EvmCode::from(bytecode);
+        assert!(eq_eip7702_version_addr(&evmcode, &addr, EIP7702_VERSION));
+
+        // Assert that From<Bytecode> preserves version.
+        let new_version = 1;
+        let mut eip_bytecode = Eip7702Bytecode::new_raw(bytes.into()).unwrap();
```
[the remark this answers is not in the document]
EdwardJES: great catch

src/storage.rs
```diff
@@ -193,3 +228,78 @@ mod in_memory;
...
+
+#[cfg(test)]
+mod tests {
+    use alloy_primitives::bytes;
+    use revm::primitives::eip7702::EIP7702_VERSION;
+
+    use super::*;
+
+    #[test]
+    fn evmcode_from_revm_bytecode_legacy() {
```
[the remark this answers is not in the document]
EdwardJES: Good idea. 

I've decided to lean on the side of caution here and do this after each `eq_bytecodes`, even though they will all fall to the same match arm. Feels safer going with this approach for evolvability and rigour. 

The alternative is to only do it once in `legacy_bytecodes` and `eip7702_bytecodes`.

Obvious downside with my approach is the like of in-lining and less terse code. 

WDYT?

src/storage.rs
```diff
@@ -193,3 +228,78 @@ mod in_memory;
...
+        assert!(eq_bytecodes(&analyzed, &evmcode));
+    }
+
+    #[test]
+    fn evmcode_from_revm_bytecode_eip7702() {
+        let addr = Address::new([0x01; 20]);
+
+        // New from address.
+        let bytecode = Bytecode::Eip7702(Eip7702Bytecode::new(addr));
+        let evmcode = EvmCode::from(bytecode.clone());
```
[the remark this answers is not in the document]
EdwardJES: Under the assumption that `assert_equals!` is used after each `eq_bytecodes`, I've not inlined here, as I felt it looked bit unwieldy if I did. 

```
 assert!(eq_bytecodes(
            &Bytecode::Eip7702(Eip7702Bytecode::new(delegated_address)),
            &EvmCode::from(Bytecode::Eip7702(Eip7702Bytecode::new(delegated_address)))
        ));
        assert_eq!(
            Bytecode::Eip7702(Eip7702Bytecode::new(delegated_address)),
            Bytecode::from(EvmCode::from(Bytecode::Eip7702(Eip7702Bytecode::new(
                delegated_address
            ))))
        );
 ```
 
 As opposed to:
 
 ```
        let bytecode = Bytecode::Eip7702(Eip7702Bytecode::new(delegated_address));
        let evm_code = EvmCode::from(Bytecode::Eip7702(Eip7702Bytecode::new(delegated_address)));
        assert!(eq_bytecodes(&bytecode, &evm_code));
        assert_eq!(bytecode, Bytecode::from(evm_code));
```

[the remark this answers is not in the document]
EdwardJES: Happy to revert to inlining if you prefer only doing `assert_eq!` once per call to `eq_bytecodes`

src/storage.rs
```diff
@@ -193,3 +227,74 @@ mod in_memory;
...
+
+#[cfg(test)]
+mod tests {
+    use alloy_primitives::bytes;
+    use revm::primitives::eip7702::EIP7702_VERSION;
+
+    use super::*;
+
+    // Bytecode from Forge's default Counter.sol contract, compiled with solc 0.8.13.
+    // https://github.com/foundry-rs/foundry/blob/nightly-fe2acca4e379793539db80e032d76ffe0110298b/testdata/multi-version/Counter.sol
```
hai-rise, line 239: This is much better, thanks!

src/storage.rs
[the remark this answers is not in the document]
hai-rise: It is much better now, thanks!

src/storage.rs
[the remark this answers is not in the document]
hai-rise: This works for me too, it also helps scope each "sub-section" well :).

src/storage.rs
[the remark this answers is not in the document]
hai-rise: Yes, it's better to reuse `evm_code` in the two asserts now!

hai-rise (approved):
Beautiful!

PR #361, opened 2024-09-24
feat: prioritize heavy txs

kien-rise:
## Before
```
Average: x2.04
```

## After

```
Average: x2.09
Max: x4.57
Min: x0.86
```

## Idea

Instead of running 12 threads, we use 8 threads for regular txs and 8 threads for heavy txs (priority txs).

The "regular" threads run as normal (possibly execute priority txs).

The "priority" threads only find and execute the priority txs. They stop as soon as there are no more priority txs.

### DIFF
--- a/benches/gigagas.rs
+++ b/benches/gigagas.rs
@@ -4,13 +4,11 @@
 
 #![allow(missing_docs)]
 
-use std::{num::NonZeroUsize, thread};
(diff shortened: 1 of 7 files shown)


kien-rise:
moved to https://github.com/risechain/pevm/pull/363

Issue #321, opened 2024-09-04
RPC Storage: better async handling


hai-rise:
https://github.com/risechain/pevm/pull/362#issuecomment-2374839062

https://github.com/risechain/pevm/blob/e6b1eb9aaae317fa2ba14ebfa501fc599cac0b75/crates/pevm/src/storage/rpc.rs#L43-L44

https://github.com/risechain/pevm/blob/e6b1eb9aaae317fa2ba14ebfa501fc599cac0b75/crates/pevm/tests/mainnet.rs#L12

https://github.com/risechain/pevm/blob/e6b1eb9aaae317fa2ba14ebfa501fc599cac0b75/bins/fetch/src/main.rs#L30

PR #362, opened 2024-09-25
chore: use tokio::test for async test

kamiyaa:
Wanted to fix this TODO, but it looks like `RpcStorage` uses a runtime, which causes this to fail?

Any thoughts on how to proceed with this? I can convert `RpcStorage` methods to be async :thinking:


hai-rise:
Linking the relevant issue here: https://github.com/risechain/pevm/issues/321 🙏.

hai-rise:
Done in #424 🙏.

PR #363, opened 2024-09-27
feat: precompute priority (heavy) txs

kien-rise:


hai-rise:
Closing for now, but still an interesting idea to consider in the new redesign.

Issue #316, opened 2024-09-04
Support EOF


hai-rise:
https://github.com/risechain/pevm/blob/24266be8d317bdcbe8f8ac10c6ce34ae7298802d/src/storage.rs#L104

hai-rise:
@byhashdong Not yet, but it's currently blocked by #358. I'll ping you when this is ready :pray:.

byhashdong:
OK, Thanks a lot!

hai-rise:
@byhashdong This is unblocked now :pray:.

byhashdong:
@hai-rise I will take it.

hai-rise:
@byhashdong Thank you!!

byhashdong:
@hai-rise Can you help clarify why do we introduce the `EVMCode` struct, but not reuse `ByteCode` from `revm`?

hai-rise:
@byhashdong This was a hard decision but `revm` types weren't optimal for `pevm`. We need to make a fork to make it easier to implement `pevm` and optimize several hot paths. Eventually, we decided on our types (`EvmCode` included) for:
- Moving `code` and `code_hash` out of `AccountBasic` (counterpart of `AccountInfo`) so we don't need to duplicate codes in the MV memory.
- Add more features like `Storage::code_hash` and `Storage::has_storage`.
- Do not persist extra data like the `raw` bytecode in `Bytecode::Eip7702`.
- (TODO) Do not duplicate bytecodes in EVM state transitions.
- (TODO) Make more key types like `Evm` thread-safe, to be pooled and shared among worker threads instead of expensive construction per transaction execution.
- Improve compatibility so library users don't need to use our forked `revm` (with the cost of some runtime type conversions).
 
For state-of-the-art performance, we'll likely need to:
- Rewrite large portions of `revm` (so cannot upstream everything), or 
- Switch to a more performant EVM executor like `evmone` or JIT/AOH compiler, or 
- Inline our EVM implementation specifically designed for parallel execution (thread-safe types and states, more native caches and instrumentation, etc.).  

And library users will only use `pevm` and `alloy` types to interface (for minimal conversions).

Nevertheless, we still provide a branch that uses `revm` instead of our custom types for integration with existing `revm` users like upstream Reth: https://github.com/risechain/pevm/pull/360/files.

byhashdong:
Nice, Thanks for you clear explain @hai-rise

byhashdong:
[...]

PR #374, opened 2024-10-07
feat: support eof(#316)

byhashdong:
We only persist `eofheader` and `eofbody` but not `raw`.

### DIFF
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -6,7 +6,6 @@ use bitvec::vec::BitVec;
 use revm::{
     interpreter::analysis::to_analysed,
     primitives::{
-        eof::{EofBody, EofHeader},
         Account, AccountInfo, Bytecode, Eip7702Bytecode, Eof, JumpTable, EIP7702_MAGIC_BYTES,
         KECCAK_EMPTY,
     },
@@ -95,12 +94,7 @@ pub struct Eip7702Code {
 
 /// EVM Object Format (EOF) code.
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
-pub struct EofCode {
-    /// EOF header.
-    pub header: EofHeader,
-    /// EOF body.
-    pub body: EofBody,
-}
+pub struct EofCode(Bytes);
 
 /// EVM Code, currently mapping to REVM's [ByteCode].
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
@@ -116,16 +110,9 @@ pub enum EvmCode {
 impl From<EvmCode> for Bytecode {
     fn from(code: EvmCode) -> Self {
         match code {
-            EvmCode::Eof(eof_code) => {
-                let mut raw = Vec::new();
-                eof_code.header.encode(&mut raw);
-                eof_code.body.encode(&mut raw);
-                Bytecode::Eof(Arc::new(Eof {
-                    header: eof_code.header,
-                    body: eof_code.body,
-                    raw: raw.into(),
-                }))
-            }
+            EvmCode::Legacy(code) => unsafe {
+                Bytecode::new_analyzed(code.bytecode, code.original_len, JumpTable(code.jump_table))
+            },
             EvmCode::Eip7702(code) => {
                 let mut raw = EIP7702_MAGIC_BYTES.to_vec();
                 raw.push(code.version);
@@ -136,9 +123,9 @@ impl From<EvmCode> for Bytecode {
                     raw: raw.into(),
                 })
             }
-            EvmCode::Legacy(code) => unsafe {
-                Bytecode::new_analyzed(code.bytecode, code.original_len, JumpTable(code.jump_table))
-            },
+            EvmCode::Eof(eof_code) => Eof::decode(eof_code.0.clone())
+                .map(|eof| Bytecode::Eof(Arc::new(eof)))
+                .unwrap_or_else(|_| Bytecode::Eof(Arc::new(Eof::default()))),
         }
     }
 }
@@ -157,10 +144,7 @@ impl From<Bytecode> for EvmCode {
                 delegated_address: code.delegated_address,
                 version: code.version,
             }),
-            Bytecode::Eof(code) => EvmCode::Eof(EofCode {
-                header: code.header.clone(),
-                body: code.body.clone(),
-            }),
+            Bytecode::Eof(code) => EvmCode::Eof(EofCode(code.raw.clone())),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/storage.rs
```diff
@@ -92,19 +93,39 @@ pub struct Eip7702Code {
     version: u8,
 }
 
+/// EVM Object Format (EOF) code.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct EofCode {
+    /// EOF header.
+    pub header: EofHeader,
```
hai-rise, line 100: I haven't looked deeply into EOF, but looking at `Eof::new` and `EofBody::into_eof`, the header can be inferred from the body? If so we can make this `EofCode` struct equivalent to `revm::EofBody`.

src/storage.rs
```diff
@@ -92,19 +93,39 @@ pub struct Eip7702Code {
...
     /// Maps delegated EIP7702 bytecode.
     Eip7702(Eip7702Code),
+    /// Maps EOF bytecode.
+    Eof(EofCode),
 }
 
 impl From<EvmCode> for Bytecode {
     fn from(code: EvmCode) -> Self {
         match code {
+            EvmCode::Eof(eof_code) => {
```
hai-rise, line 119: I think we should construct a `revm::EofBody` from our `EofCode` then call `revm::Eof::new` or `revm::EofBody::into_eof`.

Neat but can you also help re-order the arms here to `Legacy` -> `Eip7702` -> `Eof`?

src/storage.rs
```diff
@@ -238,6 +262,8 @@ mod tests {
     // Bytecode from Forge's default Counter.sol contract, compiled with solc 0.8.13.
     // https://github.com/foundry-rs/foundry/blob/nightly-fe2acca4e379793539db80e032d76ffe0110298b/testdata/multi-version/Counter.sol
     const BYTECODE: alloy_primitives::Bytes = bytes!("608060405234801561001057600080fd5b5060f78061001f6000396000f3fe6080604052348015600f57600080fd5b5060043610603c5760003560e01c80633fb5c1cb1460415780638381f58a146053578063d09de08a14606d575b600080fd5b6051604c3660046083565b600055565b005b605b60005481565b60405190815260200160405180910390f35b6051600080549080607c83609b565b9190505550565b600060208284031215609457600080fd5b5035919050565b60006001820160ba57634e487b7160e01b600052601160045260246000fd5b506001019056fea264697066735822122012c25f3d90606133b37330bf079a425dbc650fd21060dee49f715d37d97cb58f64736f6c634300080d0033");
+    const EOF_BYTECODE: alloy_primitives::Bytes =
```
hai-rise, line 265: Let's add an empty line between the `const`s!

hai-rise: Let's also add `Bytes` to the `use alloy_primitives::bytes;` above to make this concise enough for one line. Can you also add a comment on what this bytecode is?

src/storage.rs
```diff
@@ -249,6 +275,9 @@ mod tests {
             (Bytecode::Eip7702(revm), EvmCode::Eip7702(pevm)) => {
                 revm.delegated_address == pevm.delegated_address && revm.version == pevm.version
             }
+            (Bytecode::Eof(revm), EvmCode::Eof(pevm)) => {
+                revm.header == pevm.header && revm.body == pevm.body
```
hai-rise, line 279: This should be comparing our `EofCode` fields with the `revm::EofBody` fields.

src/storage.rs
```diff
@@ -297,4 +326,12 @@ mod tests {
         assert!(eq_bytecodes(&bytecode, &evm_code));
         assert_eq!(bytecode, Bytecode::from(evm_code));
     }
+
+    #[test]
+    fn eof_bytecodes() {
+        let bytecode = Bytecode::Eof(Arc::new(Eof::decode(EOF_BYTECODE).unwrap()));
+        let evm_code = EvmCode::from(bytecode.clone());
+        assert!(eq_bytecodes(&bytecode, &evm_code));
+        assert_eq!(EOF_BYTECODE, Bytecode::from(evm_code).original_bytes());
```
hai-rise, line 335: Perhaps `assert_eq!(bytecode, Bytecode::from(evm_code));` is tighter as it enforces us to convert back to `Bytecode::Eof` instead of something else that happens to has the same original bytes?

hai-rise (changes requested):
I've left a few comments. Thank you for the effort and let's hope we'll have an inline EVM implementation eventually to justify all these storage works :pray:.

src/storage.rs
```diff
@@ -92,19 +93,39 @@ pub struct Eip7702Code {
     version: u8,
 }
 
+/// EVM Object Format (EOF) code.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct EofCode {
+    /// EOF header.
+    pub header: EofHeader,
+    /// EOF body.
+    pub body: EofBody,
```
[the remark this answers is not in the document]
byhashdong: I am not very clear shall we use any filed of `EofHeader/EofBody` in pevm， as my understanding, we just save the contract code in mv_memory. If we don't use any fileds of this struct, then can we save the raw code directly?

src/storage.rs
byhashdong: Yes, `EofHeader` can be conducted from `EofBody`; but I am not very sure if we will use fields of EofHeader later.

src/storage.rs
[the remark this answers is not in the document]
hai-rise: That's a sharp insight! This is a tradeoff as making our structs closer to `revm` would make conversions faster -- also why we don't store raw bytes and re-analyze on conversion for legacy bytecodes. If decoding raw bytes is fast enough for EOF, we can store raw bytes here, and optimize later if profiling shows that it's a non-trivial bottleneck.

src/storage.rs
hai-rise: Not directly. For now, we're just converting these types to `revm` types at runtime.

src/storage.rs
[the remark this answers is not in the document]
hai-rise: Since EOF bytecodes are not yet popular, I'm happy to store the raw bytes here and optimise the conversion in the future, once EOF gains more traction 🙏.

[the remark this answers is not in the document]
byhashdong: @hai-rise, Thanks for your quick reply, will refactor soon!

src/storage.rs
```diff
@@ -92,19 +92,27 @@ pub struct Eip7702Code {
     version: u8,
 }
 
+/// EVM Object Format (EOF) code.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct EofCode(Bytes);
```
hai-rise, line 97: For simplicity let's inline this for now.

src/storage.rs
```diff
@@ -92,19 +92,27 @@ pub struct Eip7702Code {
...
 /// EVM Code, currently mapping to REVM's [ByteCode].
-// TODO: Support EOF
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub enum EvmCode {
     /// Maps both analyzed and non-analyzed REVM legacy bytecode.
     Legacy(LegacyCode),
     /// Maps delegated EIP7702 bytecode.
     Eip7702(Eip7702Code),
+    /// Maps EOF bytecode.
+    Eof(EofCode),
```
hai-rise, line 107: `Eof(Bytes)` is fine.

src/storage.rs
```diff
@@ -115,9 +123,9 @@ impl From<EvmCode> for Bytecode {
                     raw: raw.into(),
                 })
             }
-            EvmCode::Legacy(code) => unsafe {
-                Bytecode::new_analyzed(code.bytecode, code.original_len, JumpTable(code.jump_table))
-            },
+            EvmCode::Eof(eof_code) => Eof::decode(eof_code.0.clone())
```
hai-rise, line 126: We don't need to clone here, just consume the `eof_code`.

src/storage.rs
```diff
@@ -115,9 +123,9 @@ impl From<EvmCode> for Bytecode {
...
+                .map(|eof| Bytecode::Eof(Arc::new(eof)))
+                .unwrap_or_else(|_| Bytecode::Eof(Arc::new(Eof::default()))),
```
hai-rise, line 128: I'm not a fan of implicit corner logic like this. Let's panic with a `// TODO: Rewrite as [TryFrom]` added before `impl From` (also covers the `unsafe` in the `Legacy` arm).

src/storage.rs
```diff
@@ -136,7 +144,7 @@ impl From<Bytecode> for EvmCode {
                 delegated_address: code.delegated_address,
                 version: code.version,
             }),
-            Bytecode::Eof(_) => unimplemented!("TODO: Support EOF"),
+            Bytecode::Eof(code) => EvmCode::Eof(EofCode(code.raw.clone())),
```
hai-rise, line 147: We don't need to clone here, can write this as `Arc::unwrap_or_clone(code).raw`.

hai-rise (changes requested):
Almost there, love the direction already!

src/storage.rs
hai-rise: @byhashdong Last nit but this should be an `unwrap` now :pray:.

byhashdong: Ok, Got it!

hai-rise (approved):
Very clean solution, thank you!

Issue #377, opened 2024-10-09
Exit worker threads early when there are no tasks after a while


hai-rise:

hai-rise:
@byhashdong Here it is, sir!

byhashdong:
- A thread pool (of parked threads) may work better. Early experiments with rayon were very poor, so we may need to roll a specialized implementation.
  - after trying rayon, get same result, the performance is worse than original thread pool implementaion
- We should start post-processing as soon as the scheduler flags completion, instead of waiting for the scoped threads to join (very sad tail-end latency).
   - having a little test, found time gap between all threads completed and scoped completed normally less than 100us.
- We should re-consider exiting idle threads early like having no new work after 100 iterations.

hai-rise:
> after trying rayon, get same result, the performance is worse than original thread pool implementation

Hopefully one day we understand why! 

> having a little test, found time gap between all threads completed and scoped completed normally less than 100us.

Ah yes, the scope should end immediately when all threads join. The original idea is more about starting to post-process when `Scheduler` announces that execution has been done, so even before all threads have completed joining!

> confused, while task.is_some() { as the code, thread should exit if there is no new work to do.

While looking for a `Scheduler::next_task`, we currently loop until execution is done or aborted:
https://github.com/risechain/pevm/blob/bed6f220c343d03e6bef0dc7540c531444450d57/src/scheduler.rs#L118-L119

We can add a counter here to return a `None` to join a thread early when it fails to find a job after X iterations.

byhashdong:
Do you mean that still keep `thread::yield_now()` but add a counter to make sure the thread not live too long without any task to execute?

hai-rise:
@byhashdong Yes, something like that. 

FYI, `yield_now` was better than `spin_loop` and much better than not doing either (aggressive loop that may cause congestion at the atomics within) on my x86, but it didn't yield much difference on ARM. Hopefully we will know the optimal strategy for each input block, and how to play nice with the OS scheduler!

byhashdong:
Adding some stats code and running `cargo bench --bench mainnet` to do benchmark test, reporter as below:
- total `2153984` threads launched
- distribution of cycles as below:
  | sec| cycles of yields  |
  | --- | --- |
  | < 50 | 616820 |
  | 50 ~ 100 | 99462 |
  | 100 ~ 200| 217636|  
  | 200~ 500| 435890|
  | > 500 | 593066 |
[...]

PR #385, opened 2024-10-18
optimize: fine tuning thread yielding counts(#377)

byhashdong:
Thread yielding counter optimization(#377)

### DIFF
--- a/src/pevm.rs
+++ b/src/pevm.rs
@@ -172,8 +172,7 @@ impl Pevm {
         thread::scope(|scope| {
             for _ in 0..concurrency_level.into() {
                 scope.spawn(|| {
-                    let mut task_none_counter = 0;
-                    let mut task = scheduler.next_task(&mut task_none_counter);
+                    let mut task = scheduler.next_task();
                     while task.is_some() {
                         task = match task.unwrap() {
                             Task::Execution(tx_version) => {
@@ -195,7 +194,7 @@ impl Pevm {
                         }
 
                         if task.is_none() {
-                            task = scheduler.next_task(&mut task_none_counter);
+                            task = scheduler.next_task();
                         }
                     }
                 });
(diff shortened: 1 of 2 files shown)


src/pevm.rs
```diff
@@ -172,7 +172,8 @@ impl Pevm {
         thread::scope(|scope| {
             for _ in 0..concurrency_level.into() {
                 scope.spawn(|| {
-                    let mut task = scheduler.next_task();
+                    let mut task_none_counter = 0;
```
hai-rise, line 175: We can move this to the beginning of `next_task`, right before `while !self.aborted.load(Ordering::Acquire) {`?

src/scheduler.rs
```diff
@@ -161,6 +164,7 @@ impl Scheduler {
                 return Some(Task::Execution(tx_version));
             }
         }
+
```
hai-rise, line 167: Can remove this no-op.

src/scheduler.rs
```diff
@@ -105,16 +105,19 @@ impl Scheduler {
...
             if execution_idx >= self.block_size && validation_idx >= self.block_size {
-                if self.num_validated.load(Ordering::Acquire)
-                    >= self.block_size - self.min_validation_idx.load(Ordering::Acquire)
+                if *task_none_counter >= 500
+                    || (self.num_validated.load(Ordering::Acquire)
+                        >= self.block_size - self.min_validation_idx.load(Ordering::Acquire))
                 {
                     break;
                 }
+
```
hai-rise, line 119: Can remove this no-op.

hai-rise (changes requested):
Good start! I'll also benchmark with different numbers over the weekend 🙏.

hai-rise:
We'll redesign the scheduler, likely without this yield anymore (can go do state root calculation tasks instead), so closing for now 🙏.

PR #392, opened 2024-10-22
chore: store FxBuildHasher by value

DaniPopes:
`FxBuildHasher` is a ZST, there is no benefit to storing it by reference; even if it was something like a random state it would hold 1 or 2 u64s, which should still be copied rather than passed by reference


hai-rise (approved):
Good catch! We may as well use `FxBuildHasher` as a deterministic hasher for max performance. Then audit if malicious users can hand roll transactions with storage slots that cause a lot of collisions in `pevm` at a later stage :thinking:.

PR #398, opened 2024-10-29
bench: Relax the gas limit for mock transactions in gigagas bench

nekomoto911:
In gigagas bench, many transactions do not complete but instead exit with an OutOfGas error, which may lead to inaccurate bench results. For raw transfers, the issue is due to the duplication of mock account addresses and built-in precompiled contract account addresses, resulting in OutOfGas errors or errors in precompiles. ERC20 transfers and Uniswap swaps are failing midway due to the gas limit being set too low, causing most transactions to exit due to OutOfGas. We have found that the gas limit needs to be set larger than the actual gas used for some transactions to complete successfully. This is because during the execution process of the revm interpreter, the gas consumed is actually equal to gas used + gas refunded, and when the contract code uses the sstore instruction, it is required by https://eips.ethereum.org/EIPS/eip-1706 that the gas left must be >= stipend (2300) when running this instruction.
In this PR, we have relaxed the gas limit by referencing the gas expenses required for the complete execution of the transaction, so that each transaction in the bench can successfully complete.

### DIFF
--- a/benches/gigagas.rs
+++ b/benches/gigagas.rs
@@ -64,13 +64,23 @@ pub fn bench(c: &mut Criterion, name: &str, storage: InMemoryStorage, txs: Vec<T
 
 pub fn bench_raw_transfers(c: &mut Criterion) {
     let block_size = (GIGA_GAS as f64 / common::RAW_TRANSFER_GAS_LIMIT as f64).ceil() as usize;
+    // Skip the built-in precompiled contracts addresses.
+    const START_ADDRESS: usize = 1000;
+    const MINER_ADDRESS: usize = 0;
+    let storage = InMemoryStorage::new(
+        std::iter::once(MINER_ADDRESS)
+            .chain(START_ADDRESS..START_ADDRESS + block_size)
+            .map(common::mock_account),
+        None,
+        [],
+    );
     bench(
         c,
         "Independent Raw Transfers",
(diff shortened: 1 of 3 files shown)


benches/gigagas.rs
```diff
@@ -64,13 +64,23 @@ pub fn bench(c: &mut Criterion, name: &str, storage: InMemoryStorage, txs: Vec<T
 
 pub fn bench_raw_transfers(c: &mut Criterion) {
     let block_size = (GIGA_GAS as f64 / common::RAW_TRANSFER_GAS_LIMIT as f64).ceil() as usize;
+    // Skip the built-in precompiled contracts addresses.
```
hai-rise, line 67: The effect is minor but love the clarity!

hai-rise (approved):
Amazing work! 

We were aware of the precompiles in raw transfers (due to hash collisions when playing with `SuffixHasher` 😅) but it was so minor we kept it on.

However, the effect on ERC-20 & Uniswap is massive! Speedup went up for ERC-20 but down for Uniswap. Also makes the peak raw execution speed down from 55 Ggas/s to only 30 Ggas/s.

We did face failed Uniswap transactions when benchmarking pevm integration into the sequencer (due to deadline config actually), but never got time to re-inspect this benchmark. Thank you a ton for this contribution 🙏.

Issue #397, opened 2024-10-28
Return error instead of panicking for mismatched nonce during post-processing


hai-rise:
A cleaner version of: https://github.com/kien-rise/pevm/commit/1e555cc995abef34b57129dd8d32875219cd689e.

PR #400, opened 2024-10-29
fix: do not panic in nonce check

kien-rise:
Resolves https://github.com/risechain/pevm/issues/397

### DIFF
--- a/src/pevm.rs
+++ b/src/pevm.rs
@@ -5,7 +5,7 @@ use std::{
     thread,
 };
 
-use alloy_primitives::U256;
+use alloy_primitives::{TxNonce, U256};
 use alloy_rpc_types::{Block, BlockTransactions};
 use hashbrown::HashMap;
 use revm::{
@@ -24,7 +24,7 @@ use crate::{
     vm::{
         build_evm, ExecutionError, PevmTxExecutionResult, Vm, VmExecutionError, VmExecutionResult,
     },
-    EvmAccount, MemoryEntry, MemoryLocation, MemoryValue, Storage, Task, TxVersion,
+    EvmAccount, MemoryEntry, MemoryLocation, MemoryValue, Storage, Task, TxIdx, TxVersion,
 };
 
 /// Errors when executing a block with pevm.
@@ -36,6 +36,15 @@ pub enum PevmError<C: PevmChain> {
     MissingTransactionData,
     /// Invalid input transaction.
     InvalidTransaction(C::TransactionParsingError),
+    /// Nonce too low or too high
+    NonceMismatch {
+        /// Transaction index
+        tx_idx: TxIdx,
+        /// Nonce from tx (from the very input)
+        tx_nonce: TxNonce,
+        /// Nonce from state and execution
+        executed_nonce: TxNonce,
+    },
     /// Storage error.
     // TODO: More concrete types than just an arbitrary string.
     StorageError(String),
@@ -287,10 +296,23 @@ impl Pevm {
                         _ => unreachable!(),
                     }
                     // Assert that evaluated nonce is correct when address is caller.
-                    debug_assert!(
-                        tx.caller != address || tx.nonce.map_or(true, |n| n + 1 == nonce)
-                    );
-
+                    if tx.caller == address {
+                        if let Some(tx_nonce) = tx.nonce {
+                            let executed_nonce = if nonce == 0 {
+                                return Err(PevmError::UnreachableError);
+                            } else {
+                                nonce - 1
+                            };
+                            if tx_nonce != executed_nonce {
+                                // TODO: Consider falling back to sequential instead
+                                return Err(PevmError::NonceMismatch {
+                                    tx_idx: *tx_idx,
+                                    tx_nonce,
+                                    executed_nonce,
+                                });
+                            }
+                        }
+                    }
                     // SAFETY: The multi-version data structure should not leak an index over block size.
                     let tx_result = unsafe { fully_evaluated_results.get_unchecked_mut(*tx_idx) };
                     let account = tx_result.state.entry(address).or_default();
--- a/tests/ethereum/main.rs
+++ b/tests/ethereum/main.rs
@@ -156,17 +156,8 @@ fn run_test_unit(path: &Path, unit: TestUnit) {
                 ),
             ) {
                 // EIP-2681
-                (Some("TR_NonceHasMaxValue"), Ok(exec_results)) => {
-                    assert!(exec_results.len() == 1);
-                    assert!(exec_results[0].receipt.status.coerce_status());
-                    // This is overly strict as we only need the newly created account's code to be empty.
-                    // Extracting such account is unjustified complexity so let's live with this for now.
-                    assert!(exec_results[0].state.values().all(|account| {
-                        match account {
-                            Some(account) => account.code_hash.is_none(),
-                            None => true,
-                        }
-                    }));
+                (Some("TR_NonceHasMaxValue"), Err(err)) => {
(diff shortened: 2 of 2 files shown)


src/pevm.rs
```diff
@@ -36,6 +32,13 @@ pub enum PevmError<C: PevmChain> {
     MissingTransactionData,
     /// Invalid input transaction.
     InvalidTransaction(C::TransactionParsingError),
+    /// Nonce too low or too high
+    NonceMismatch {
+        /// Nonce from tx (observed value)
```
hai-rise, line 37: This is the `expected` value? 

A more helpful error would also include the transaction index. Maybe `tx_idx`, `expected_nonce`, and `observed_nonce`.

src/pevm.rs
```diff
@@ -263,10 +266,16 @@ impl Pevm {
                         _ => unreachable!(),
                     }
                     // Assert that evaluated nonce is correct when address is caller.
-                    debug_assert!(
-                        tx.caller != address || tx.nonce.map_or(true, |n| n + 1 == info.nonce)
-                    );
-
+                    if tx.caller == address {
+                        if let Some(nonce) = tx.nonce {
```
hai-rise, line 270: `Some(tx_nonce)` or `Some(expected_nonce)` is clearer.

src/pevm.rs
```diff
@@ -263,10 +266,16 @@ impl Pevm {
...
+                            if nonce + 1 != info.nonce {
```
hai-rise, line 271: For consistency can do `if info.nonce - 1 != expected_nonce ` instead of `+ 1` here and `- 1` below. Alternatively can invest in an intermediate `let observed_nonce = info.nonce - 1`.

src/pevm.rs
```diff
@@ -263,10 +266,16 @@ impl Pevm {
...
+                                return Err(PevmError::NonceMismatch {
```
hai-rise, line 272: Let's add a `// TODO: Consider falling back to sequential instead`

hai-rise (changes requested):
This was supposed to be a warmup exercise for Cesar but yeah, I've left a few nits.

src/pevm.rs
kien-rise: The struct is "copied" from revm: https://github.com/risechain/revm/blob/main/crates/primitives/src/result.rs#L278-L285

```rust
    NonceTooHigh {
        tx: u64,
        state: u64,
    },
    NonceTooLow {
        tx: u64,
        state: u64,
    },
```

While adding `tx_idx` is good, using names `expected_nonce` and `observed_nonce` might be ambiguous. To me, `tx` is the "observed" value while `state` is the "expected" value. If this error happens, that's because the input is invalid (e.g. the block building service is feeding a nonce-too-low tx or nonce-to-high tx), not because the execution or state is wrong.

src/pevm.rs
kien-rise: It should be `Some(observed_nonce)`, rather than `Some(expected_nonce)`.

To avoid ambiguity, let's just use `Some(tx_nonce)`.

src/pevm.rs
hai-rise: Yeah, the problem is that either side can be wrong. A race condition in `pevm` or a corrupted state during canonical syncing would make `tx` the expected value instead. I got confused because `observed` is typically a calculated value. 

So yeah, let's do `tx_idx`, `tx_nonce`, and `calculated_nonce`/`executed_nonce`, and let the caller decide which is wrong.

src/pevm.rs
hai-rise: Yeah, let's go with `tx_nonce`.

src/pevm.rs
kien-rise: If `PevmError::NonceMismatch` happens, it is likely that the input is invalid (i.e. because the block building service is feeding pevm with a nonce-too-low or nonce-too-high tx).

So I think we should not "consider falling back to sequential". We should `return Err(...)` instead.

src/pevm.rs
kien-rise: I will change to `info.nonce.saturating_sub(1)`. Although it's likely that `info.nonce > 0`, I cannot prove that. Note that the expression `0u64 - 1` will cause a panic.

src/pevm.rs
hai-rise: No, block building shouldn't wait until post-processing to arrive at an error that needs to rework completely. Nonce checking should be done before or during block building instead. On the other hand, nonce errors during post-processing is possible if there's a bug in lazy evaluation or some weird race conditions that we haven't addressed.

src/pevm.rs
hai-rise: Release builds will wrap instead of panicking, but yeah, it's best practice to use `saturating_x` for numerical operations like this.

src/pevm.rs
kien-rise: resolved

src/pevm.rs
kien-rise: comment added

src/pevm.rs
kien-rise: changed to `.saturating_sub(1)`

hai-rise: Just had another look and `saturating_x` is actually not the solution? It still allows `tx_nonce == 0` and `nonce == 0` to pass the check. The best solution to not assume the correctness of other components is to check if `nonce` is zero first and return `UnreachableError` (or something better) if it is. Then `-` would satisfy our check.

kien-rise: > Just had another look and `saturating_x` is actually not the solution? It still allows `tx_nonce == 0` and `nonce == 0` to pass the check. The best solution to not assume the correctness of other components is to check if `nonce` is zero first and return `UnreachableError` (or something better) if it is. Then `-` would satisfy our check.

Is this ok?

```
// Assert that evaluated nonce is correct when address is caller.
if tx.caller == address {
    if let Some(tx_nonce) = tx.nonce {
        if info.nonce == 0 {
            return Err(PevmError::UnreachableError);
        }
        if tx_nonce != info.nonce - 1 {
            // TODO: Consider falling back to sequential instead
            return Err(PevmError::NonceMismatch {
                tx_idx: *tx_idx,
                tx_nonce,
                executed_nonce: info.nonce - 1,
            });
        }
    }
}
```

------------------

*Another idea*: Can I use `.wrapping_sub(1)`? It's not hard to decipher `NonceMismatch { tx_nonce: 0, executed_nonce: 18446744073709551615 }` though.

hai-rise: Maybe this would add more context to the check:
```rust
let executed_nonce = if nonce == 0 {
    return Err(PevmError::UnreachableError);
} else {
    nonce - 1
};
```

Or we can move the `UnreachableError` check to right after the `match` on `MemoryEntry`, and only proceed to `if tx.caller == address {` if that check passes.

kien-rise:
all discussions resolved

kien-rise:
Sorry for the CI failure, I have fixed the test at `tests/ethereum/main.rs`.

hai-rise (approved):
The test simplification is nice. We probably want a dedicated error message for this max case but yeah, can do later on.

cr-rise:
@hai-rise the code is different. It is not failing if `tx.caller != address `, as it was before :

```rust
debug_assert!(
    tx.caller != address || tx.nonce.map_or(true, |n| n + 1 == nonce)
);
```

cr-rise:
@hai-rise I got it, I was misreading somehow. The PR is correct.

hai-rise:
@cr-rise No worries at all. I just made a boolean mistake today too! https://github.com/risechain/pevm/pull/415#discussion_r1825715235

PR #402, opened 2024-10-30
Split binary and library to separate crates and make RPC feature

### LINKED ISSUE #298: Put RPC Storage under a dedicated feature
Add an `rpc-storage` feature that enables this module:
https://github.com/risechain/pevm/blob/7c9529e0f5435776c5f25af384d955e6dfdefb50/src/storage/rpc.rs#L1-L2

And these dependencies:
https://github.com/risechain/pevm/blob/7c9529e0f5435776c5f25af384d955e6dfdefb50/Cargo.toml#L43-L50

i1i1:
Closes #298


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "tests/ethereum/tests"]
-	path = tests/ethereum/tests
+	path = pevm/tests/ethereum/tests
```
hai-rise, line 2: We also need to update the `working-directory` for lint and tests in CI, which would actually fail currently.

Cargo.toml
```diff
@@ -1,89 +1,62 @@
-[package]
```
hai-rise, line 1: Nit but let's format TOML files as well for consistency.

Cargo.toml
```diff
@@ -1,89 +1,62 @@
...
-[features]
-optimism = [
-    "revm/optimism",
-    "dep:op-alloy-consensus",
-    "dep:op-alloy-network",
-    "dep:op-alloy-rpc-types",
+[workspace]
+default-members = ["bins/*"]
+members = [
+  "bins/*",
```
hai-rise, line 4: Nit but I prefer to list all members instead of listing patterns.

Cargo.toml
```diff
@@ -1,89 +1,62 @@
...
-edition = "2021"
-
-[features]
-optimism = [
-    "revm/optimism",
-    "dep:op-alloy-consensus",
-    "dep:op-alloy-network",
-    "dep:op-alloy-rpc-types",
+[workspace]
+default-members = ["bins/*"]
```
hai-rise, line 2: This should be the core `pevm` library?

bins/fetch/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "pevm-fetch"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+bincode.workspace = true
+clap = { workspace = true, features = ["derive"] }
+flate2.workspace = true
+serde_json = { workspace = true, features = ["preserve_order"] }
```
hai-rise, line 10: Good features all crates enable by default can be better declared in the workspace `Cargo.toml`?

Cargo.toml
```diff
@@ -1,89 +1,62 @@
...
-[[bench]]
-name = "gigagas"
-harness = false
-
-[profile.release]
-codegen-units = 1
-panic = "abort"
-lto = "fat"
+tokio = "1.40.0"
+walkdir = "2.5.0"
```
hai-rise, line 62: Nit but let's add final trailing newlines :pray:.

pevm/Cargo.toml
```diff
@@ -0,0 +1,80 @@
+[package]
+name = "pevm"
+version = "0.1.0"
+edition = "2021"
+
+[features]
+defaults = []
+rpc = [
+  "tokio",
```
hai-rise, line 9: These should be `dep:*`? Nit but let's also sort the deps alphabetically and add a newline to separate each feature set.

pevm/src/lib.rs
```diff
@@ -221,9 +221,11 @@ mod pevm;
 pub use pevm::{execute_revm_sequential, Pevm, PevmError, PevmResult};
 mod scheduler;
 mod storage;
+#[cfg(feature = "rpc")]
```
hai-rise, line 224: Nit but I'd put it below the default list.

hai-rise (changes requested):
I have left a few nits. We also need to make the CI lint & tests `--workspace` and fix accordingly (it's failing).

`pevm/pevm` is a bit uncomfortable but overall I like this direction. One day we may have another library crate to make it `pevm/crates/{pevm, ???}`, too :).

bins/fetch/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "pevm-fetch"
```
hai-rise, line 2: I like this name better. Let's also update it in the README.

pevm/Cargo.toml
```diff
@@ -0,0 +1,80 @@
+[package]
```
hai-rise, line 1: Let's add a `crates` directory to make it more "standard", so this should be `pevm/crates/pevm/Cargo.toml`.

Cargo.toml
i1i1: Should I use https://github.com/tamasfe/taplo or smth else for formatting?

Cargo.toml
i1i1: Hmm, this just means that when we'll do `cargo build` it'll build binaries, but I can change it

bins/fetch/Cargo.toml
i1i1: You mean, that I should enable some features on workspace level, right? Should I add smth else, apart from `serde_json`?

Cargo.toml
hai-rise: Yes, that is a good choice. I'll add it to the upcoming `CONTRIBUTING.md`!

Cargo.toml
hai-rise: Yeah, the binary is very niche so most of time we lint, build, test, and bench the library by default (for development).

bins/fetch/Cargo.toml
hai-rise: Yes, let's port all feature choices to the workspace `Cargo.toml` for simplicity.

.github/workflows/ci.yml
```diff
@@ -25,7 +25,7 @@ jobs:
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
 
       - name: Lint
-        run: cargo clippy --all-targets --all-features
+        run: cargo clippy --all --all-targets --all-features
```
hai-rise, line 28: Let's use `--workspace`:
```
Package Selection:
  -p, --package [<SPEC>]  Package to build (see `cargo help pkgid`)
      --workspace         Build all packages in the workspace
      --exclude <SPEC>    Exclude packages from the build
      --all               Alias for --workspace (deprecated)
```

README.md
```diff
@@ -77,10 +77,10 @@ $ cargo build
 
 #### Fetch Real Blocks
```
hai-rise, line 78: Can you also update the paths in `Testing` and `Benchmarks`?

bins/fetch/Cargo.toml
hai-rise: You've only done it for `pevm`, let's do the same here.

crates/pevm/Cargo.toml
```diff
@@ -0,0 +1,76 @@
...
+]
+
+[dependencies]
+alloy-chains.workspace = true
+alloy-consensus.workspace = true
+alloy-primitives.workspace = true
+alloy-rlp.workspace = true
+alloy-rpc-types.workspace = true
+alloy-trie.workspace = true
+# We can roll our own but [revm] depends on this anyway.
```
hai-rise, line 28: Very nit but this comment is better ported to the workspace `Cargo.toml` now.

crates/pevm/tests/common/mod.rs
```diff
@@ -54,36 +54,35 @@ pub const RAW_TRANSFER_GAS_LIMIT: u64 = 21_000;
...
+        File::open(data_dir.join("bytecodes.bincode.gz")).unwrap(),
     )))
     .unwrap();
 
-    for block_path in fs::read_dir("data/blocks").unwrap() {
+    for block_path in fs::read_dir(data_dir.join("blocks")).unwrap() {
         let block_path = block_path.unwrap().path();
         let block_number = block_path.file_name().unwrap().to_str().unwrap();
 
+        let dir = data_dir.join("blocks").join(block_number);
```
hai-rise, line 69: `block_dir` for clarity :pray:.

hai-rise (changes requested):
Almost there!

crates/pevm/Cargo.toml
```diff
@@ -0,0 +1,76 @@
+[package]
+name = "pevm"
+version = "0.1.0"
+edition = "2021"
+
+[features]
+defaults = []
+rpc = [
```
hai-rise, line 8: Oh one last thing but it should be `rpc-storage` since `rpc` is very broad and can be misleading.

hai-rise (approved):
Really like the new structure, thanks! Also appreciated that you suggested further improvements over what was needed 🙏.

Issue #382, opened 2024-10-15
Implement a fine-tuned EVM instead of relying on an external one


hai-rise:
Context: https://github.com/risechain/pevm/issues/316#issuecomment-2395292657, #319, #373.

TL;DR, we want an implementation specifically designed for parallel execution with thread-safe types, states, more native caches and instrumentation, no overheads for lazy, etc. Using an external sequential-first implementation like `revm` adds too much overhead and unwanted type conversions. 

EVM isn't that complicated anyway: https://github.com/lambdaclass/lambda_ethereum_rust/tree/main/crates/vm/levm.

byhashdong:
@hai-rise I'm also very interested in this topic. 
Do you have any plan and task splits or architecture design now, or just fork from levm?

hai-rise:
@byhashdong This is an ambitious project! 

A fork is likely a bad choice as we want low-level control over EVM execution right within `pevm::Vm`, for instance, to use the multi-version memory directly instead of going through another (frictional) layer of `JournaledState` -> `EvmState` in `revm`. We may want to reuse/reference utility functions (like era-based gas calculation), but it must be very different architecturally. 

The current plan is to take baby steps like in #383. We will need many iterations before we can completely remove `revm` usage. I'll create smaller tasks along the way that you can pick if you are interested 🙏.

byhashdong:
I'm thrilled to hear this ambitious project, so eager to see your task list. 
And if any design doc could be provided, that will be more helpful for code implementation.

hai-rise:
@byhashdong We'll get there, sir 🫡.

PR #403, opened 2024-10-30
feat: downgrade deps to make it compatible to reth 1.1.0

kien-rise:
We need these changes to make sure there is only one version of `alloy-eip7702` in the reth integration. Otherwise, there will be compile errors.

```diff
- op-alloy-consensus = { version = "0.5.0", optional = true }
- op-alloy-network = { version = "0.5.0", optional = true }
- op-alloy-rpc-types = { version = "0.5.0", optional = true }
+ op-alloy-consensus = { version = "0.4.0", optional = true }
+ op-alloy-network = { version = "0.4.0", optional = true }
+ op-alloy-rpc-types = { version = "0.4.0", optional = true }
```

This change is to update `revm` to the [`reth` branch](https://github.com/risechain/revm/commits/reth):
```diff
- revm = { git = "https://github.com/risechain/revm", rev = "bb350a971e5f39a8f822ace081584a3a45fd7c6a", features = [
+ revm = { git = "https://github.com/risechain/revm", rev = "fa685fbacc56718bdcc8aa11f851e5070c2d63b3", features = [
```

The other changes are just to make the code compiles.

-----

Yes, the target branch of this PR is [reth-integration-2](https://github.com/risechain/pevm/tree/reth-integration-2) because those changes are not needed on `main`.

PS: Actually, if we want, we can keep `op-alloy-*` changes on `main` branch and `revm` change on this branch (i.e. split this PR into two PRs). We can do that to keep this branch minimal.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,7 +24,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
 dependencies = [
  "cfg-if",
- "getrandom",
  "once_cell",
  "version_check",
  "zerocopy",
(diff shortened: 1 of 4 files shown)


hai-rise (approved):

hai-rise:
> PS: Actually, if we want, we can keep op-alloy-* changes on main branch and revm change on this branch (i.e. split this PR into two PRs). We can do that to keep this branch minimal.

Nah, that's fine. Our end goal is to remove `revm` completely (#382) but keep Alloy (up-to-date) for primitive types, so this PR is in the right direction already. Once the old `reth-integration` is not needed anymore I'll squash this branch and force-push to it and keep this `-2` as the new backup.

PR #406, opened 2024-10-31
fix: handle nonce error properly

kien-rise:
If there is a nonce-too-high in the input, `pevm` will freeze at this line: https://github.com/risechain/pevm/blob/c78b772b0fb9046711fee4a1ce87dddfae180ecd/src/vm.rs#L401

Previously, there was a quick-and-dirty fix here: https://github.com/kien-rise/pevm/commit/bc95f3fb8cb57b4611866a4cf45e58daa011fb1c.

This PR provides the proper fix.

### DIFF
--- a/src/vm.rs
+++ b/src/vm.rs
@@ -1,6 +1,8 @@
+use std::cmp::Ordering;
+
 use alloy_primitives::TxKind;
 use alloy_rpc_types::Receipt;
-use hashbrown::HashMap;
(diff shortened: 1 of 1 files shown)


kien-rise:
The current code is too messy. Let me refactor it first.

kien-rise:
moved to https://github.com/risechain/pevm/pull/408

PR #409, opened 2024-11-01
trying new formula for empty task check(#377)

byhashdong:


hai-rise:
We'll redesign the scheduler, likely without this yield anymore (can go do state root calculation tasks instead), so closing for now 🙏.

PR #415, opened 2024-11-01
Remove allocator compilation to speed up CI a bit

i1i1:


hai-rise (changes requested):
The allocator change is not very pleasant DX. How about adding a `global-alloc` feature and putting the allocators under it? One can benchmark normally or with `--features global-alloc`. For CI we can omit it from the feature list.

And let's cherry-pick the lint work into a dedicated PR.

i1i1:
Sorry for confusion! Makes sense. I just removed commit related to allocators at all. I'll make another pr!

crates/pevm/src/mv_memory.rs
```diff
@@ -98,7 +98,7 @@ impl MvMemory {
         let mut last_location_idx = 0;
         while last_location_idx < last_locations.write.len() {
             let prev_location = unsafe { last_locations.write.get_unchecked(last_location_idx) };
-            if !write_set.iter().any(|(l, _)| l == prev_location) {
+            if write_set.iter().all(|(l, _)| l != prev_location) {
```
hai-rise, line 101: `any` is preferred as it is short-circuiting and this is a common case as most transaction re-execution still writes to the same memory locations. `all` would be visibly slower for transactions that write to many locations.

Cargo.toml
```diff
@@ -8,11 +8,103 @@ codegen-units = 1
 panic = "abort"
 lto = "fat"
 
+# Copied from https://github.com/paradigmxyz/reth/blob/bd8c4eceb20c39c6e501d06cf906469329340bb9/Cargo.toml#L159
```
hai-rise, line 11: This copy-pasting from Reth is pretty noisy. Let's only keep things that are relevant for now: 
- `rust.rust_2018_idioms = "warn"` for `[workspace.lints]`, and:
```
[workspace.lints.clippy]
cloned_instead_of_copied = "warn"
derive_partial_eq_without_eq = "warn"
doc_markdown = "warn"
explicit_iter_loop = "warn"
if_not_else = "warn"
if_then_some_else_none = "warn"
match_same_arms = "warn"
missing_const_for_fn = "warn"
redundant_clone = "warn"
redundant_else = "warn"
type_repetition_in_bounds = "warn"
unnested_or_patterns = "warn"
use_self = "warn"
```

hai-rise (changes requested):
Two comments but overall really really like these improvements!

crates/pevm/src/mv_memory.rs
i1i1: I think this is the semantic equivalent, actually, and there is no performance difference as well. You can write `.all` and `.any` in terms of each other like this:
```rust
fn any(&self, pred: impl FnMut(I) -> bool) -> bool {
  !self.all(|i| !pred(i))
}
fn all(&self, pred: impl FnMut(I) -> bool) -> bool {
  !self.any(|i| !pred(i))
}
```

You can also compare implementations of those 2 in stdlib
https://doc.rust-lang.org/src/core/iter/traits/iterator.rs.html#2751-2756
https://doc.rust-lang.org/src/core/iter/traits/iterator.rs.html#2804-2811

hai-rise: Aha, should check the doc before commenting next time. No need to change then 🙏.

Cargo.toml
```diff
@@ -9,11 +9,24 @@ panic = "abort"
 lto = "fat"
 
 [workspace.lints]
-rust.missing_debug_implementations = "warn"
-rust.missing_docs = "warn"
-rust.unreachable_pub = "warn"
+rust.rust_2018_idioms = { level = "deny", priority = -1 }
```
hai-rise, line 12: Let's keep this as `"warn"` for consistency.

hai-rise (changes requested):
Last nit I promise 🥲.

Cargo.toml
```diff
@@ -9,11 +9,24 @@ panic = "abort"
 lto = "fat"
 
 [workspace.lints]
-rust.missing_debug_implementations = "warn"
```
hai-rise, line 12: Oh no, we should keep these 3 and only add `rust_2018_idioms` 😅.

hai-rise (approved):
Beautiful!!

PR #416, opened 2024-11-01
Make separate feature for allocators

i1i1:


crates/pevm/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [features]
-defaults = []
+defaults = ["rpc-storage", "optimism"]
```
hai-rise, line 7: This is cleaner for CI but would be hidden friction for typical library users. If many cannot make `serde` default then we cannot make these big ones default. 

To avoid duplication in CI and future mistakes, we can add a `full` feature that enables these 2. Like how `tokio`'s `full` enables most but not `test-util`.

hai-rise (changes requested):
Almost there!

hai-rise (approved):
Nice!

PR #418, opened 2024-11-01
doc(README): fix command

hai-rise:


i1i1 (approved):
Thanks! I dunno why I thought that `--bin` would work :sweat:

hai-rise:
@i1i1 Same here 😅!

PR #424, opened 2024-11-05
RPC Storage: better async handling

### LINKED ISSUE #321: RPC Storage: better async handling

cr-rise:
Added `block_on` to execute futures to support async functions with the sync traits for storage

Fixes #321

### DIFF
--- a/bins/fetch/src/main.rs
+++ b/bins/fetch/src/main.rs
@@ -14,7 +14,6 @@ use alloy_rpc_types::{BlockId, BlockTransactionsKind};
 use clap::Parser;
 use flate2::{bufread::GzDecoder, write::GzEncoder, Compression};
 use pevm::{
-    block_on,
     chain::{PevmChain, PevmEthereum},
     EvmAccount, EvmCode, Pevm, RpcStorage,
 };
@@ -27,10 +26,8 @@ struct Fetch {
     block_id: BlockId,
 }
 
-// TODO: async main?
-// TODO: Binary formats to save disk?
-// TODO: Test block after fetching it.
-fn main() -> Result<(), Box<dyn Error>> {
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
     let Fetch { block_id, rpc_url } = Fetch::parse();
 
     // Define provider.
@@ -40,7 +37,9 @@ fn main() -> Result<(), Box<dyn Error>> {
     );
 
     // Retrive block from provider.
-    let block = block_on(provider.get_block(block_id, BlockTransactionsKind::Full))
+    let block = provider
+        .get_block(block_id, BlockTransactionsKind::Full)
+        .await
         .map_err(|err| format!("Failed to fetch block from provider. {err}"))?
         .ok_or(format!("No block found for ID: {:?}", block_id))?;
 
--- a/bins/fetch/src/main.rs
+++ b/bins/fetch/src/main.rs
@@ -14,11 +14,11 @@ use alloy_rpc_types::{BlockId, BlockTransactionsKind};
 use clap::Parser;
 use flate2::{bufread::GzDecoder, write::GzEncoder, Compression};
 use pevm::{
+    block_on,
     chain::{PevmChain, PevmEthereum},
     EvmAccount, EvmCode, Pevm, RpcStorage,
 };
 use reqwest::Url;
-use tokio::runtime::Runtime;
 
 #[derive(Parser, Debug)]
 /// Fetch is a CLI tool to fetch a block from an RPC provider, and snapshot that block to disk.
@@ -40,9 +40,7 @@ fn main() -> Result<(), Box<dyn Error>> {
     );
 
     // Retrive block from provider.
-    let runtime = Runtime::new()?;
-    let block = runtime
-        .block_on(provider.get_block(block_id, BlockTransactionsKind::Full))
+    let block = block_on(provider.get_block(block_id, BlockTransactionsKind::Full))
         .map_err(|err| format!("Failed to fetch block from provider. {err}"))?
         .ok_or(format!("No block found for ID: {:?}", block_id))?;
 
--- a/bins/fetch/src/main.rs
+++ b/bins/fetch/src/main.rs
@@ -26,6 +26,8 @@ struct Fetch {
     block_id: BlockId,
 }
 
+// TODO: Binary formats to save disk?
+// TODO: Test block after fetching it.
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error>> {
     let Fetch { block_id, rpc_url } = Fetch::parse();
--- a/crates/pevm/Cargo.toml
+++ b/crates/pevm/Cargo.toml
@@ -5,23 +5,17 @@ edition = "2021"
 
 [features]
 defaults = []
-
 full = ["optimism", "rpc-storage"]
-
+rpc-storage = ["dep:reqwest", "dep:alloy-transport", "dep:alloy-transport-http"]
 optimism = [
   "revm/optimism",
+  "dep:alloy-transport",
+  "dep:alloy-transport-http",
   "dep:op-alloy-consensus",
   "dep:op-alloy-network",
   "dep:op-alloy-rpc-types",
+  "dep:reqwest"
 ]
-
-rpc-storage = [
-  "dep:alloy-transport",
-  "dep:alloy-transport-http",
-  "dep:reqwest",
-  "dep:tokio",
-]
-
 global-alloc = ["dep:rpmalloc", "dep:snmalloc-rs", "dep:tikv-jemallocator"]
 
 [dependencies]
@@ -53,7 +47,7 @@ op-alloy-rpc-types = { workspace = true, optional = true }
 alloy-transport = { workspace = true, optional = true }
 alloy-transport-http = { workspace = true, optional = true }
 reqwest = { workspace = true, optional = true }
-tokio = { workspace = true, optional = true }
+tokio = { workspace = true }
 
 # Various allocators for benchmarks
 rpmalloc = { workspace = true, optional = true }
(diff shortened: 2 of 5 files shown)


crates/pevm/src/lib.rs
```diff
@@ -4,14 +4,49 @@
...
+
+#[inline(always)]
+/// Awaits on a future using the current Tokio runtime if it exists,
+/// or creates a new one (at most once) if it doesn't.
+pub fn block_on<F: Future + Send>(future: F) -> F::Output
+where
+    F::Output: Send,
+{
+    // Get a handle to the current Tokio runtime
+    if Handle::try_current().is_ok() {
```
hai-rise, line 31: This implementation seems off as it doesn't reuse the existing handle and duplicates the usage of `GLOBAL_RUNTIME` in both branches. I think the intended logic must look somewhat like this:

```rust
if let Ok(handle) = Handle::try_current() {
    handle
} else {
    GLOBAL_RUNTIME
        .get_or_init(|| Runtime::new().expect("Failed to create Tokio runtime"))
        .handle()
        .clone()
}
.block_on(future)
```

crates/pevm/src/lib.rs
```diff
@@ -4,14 +4,49 @@
...
+        thread::scope(|s| {
```
hai-rise, line 32: This is a cool trick to avoid creating a new runtime that blocks in the main thread of another runtime! But I guess the more "standard" solution is to use `tokio::task::block_in_place`. It helps remove the `Send` requirement as well.

crates/pevm/src/lib.rs
```diff
@@ -4,14 +4,49 @@
...
+};
+use tokio::runtime::{Handle, Runtime};
 
 use alloy_primitives::{Address, B256, U256};
 use bitflags::bitflags;
 use hashbrown::HashMap;
 use rustc_hash::FxBuildHasher;
 use smallvec::SmallVec;
 
+static GLOBAL_RUNTIME: OnceLock<Runtime> = OnceLock::new();
```
hai-rise, line 21: I'm not a fan of global variables and `pevm` should not export something like `block_on`, so let's just promote the current `runtime` in `RpcStorage` from `Runtime` to `OnceLock<Runtime>`.

crates/pevm/src/storage/rpc.rs
```diff
@@ -51,36 +48,32 @@ impl<N: Network> RpcStorage<N> {
...
-                } else {
-                    return result;
-                }
+        let mut lives = RETRY_LIMIT;
+        let mut delay = Duration::from_millis(INITIAL_DELAY_MILLIS);
+
+        loop {
+            let result = request().await;
+            if lives > 0 && result.is_err() {
+                sleep(delay).await;
```
hai-rise, line 70: Let's quantify common names like `sleep` and `spawn` as the difference between `std` and `tokio` (or other libraries) can be confusing.

crates/pevm/src/storage/rpc.rs
```diff
@@ -51,36 +48,32 @@ impl<N: Network> RpcStorage<N> {
...
-            runtime: Runtime::new().unwrap(),
         }
     }
 
     /// Send a request and retry many times if needed.
     /// This util is made to avoid error 429 Too Many Requests
-    /// <https://en.wikipedia.org/wiki/Exponential_backoff>
-    fn fetch<T, E, R: IntoFuture<Output = Result<T, E>>>(
+    /// https://en.wikipedia.org/wiki/Exponential_backoff
+    async fn fetch<T, E, R: IntoFuture<Output = Result<T, E>>>(
```
hai-rise, line 57: Nice!

crates/pevm/tests/mainnet.rs
```diff
@@ -9,9 +9,11 @@ use pevm::chain::{PevmChain, PevmEthereum};
 pub mod common;
 
 // TODO: [tokio::test]?
-#[test]
+#[tokio::test]
```
hai-rise, line 12: Unlike `tokio::main`, the default flavour for `tokio::test` is actually single-threaded! We must add the `multi_thread` flavour here, otherwise it would:
- Panic with `tokio::task::block_in_place`.
- Deadlock when the provided runtime is used alongside the global one (like using `await` on `provider.get_block`) with `std::thread::scope`.

crates/pevm/tests/mainnet.rs
```diff
@@ -34,14 +36,12 @@ fn mainnet_blocks_from_rpc() {
...
     ] {
-        let runtime = tokio::runtime::Runtime::new().unwrap();
         let provider = ProviderBuilder::new().on_http(rpc_url.clone());
-        let block = runtime
-            .block_on(
-                provider.get_block(BlockId::number(block_number), BlockTransactionsKind::Full),
-            )
-            .unwrap()
-            .unwrap();
+        let block = block_on(
```
hai-rise, line 40: We should `await` here instead of `block_on`.

hai-rise (changes requested):
I like the bold approach! But yeah, still need more work to close this :pray:.

crates/pevm/src/lib.rs
cr-rise: It was but it was blocking, due to `tokio::test` using a single thread. I'll update `block_on` now that it works

crates/pevm/src/lib.rs
cr-rise: @hai-rise it was a hack to prevent the panics of `block_in_place` due to `tokio::test`, but I'll remove it.

hai-rise: Yeah, but it would still deadlock in a current-thread runtime if it is used to drive another asynchronous task (like fetching the full block before executing it). If anything `block_in_place`'s panic message suggests library users switch to a multi-thread runtime!

Let's settle on not supporting current-thread runtime and either:
- Build `RpcStorage`'s own multi-thread runtime if there is not an active one; or
- Expect the active runtime to be multi-thread.

bins/fetch/src/main.rs
```diff
@@ -27,10 +26,8 @@ struct Fetch {
     block_id: BlockId,
 }
 
-// TODO: async main?
-// TODO: Binary formats to save disk?
```
hai-rise, line 31: Let's keep the bottom 2 `TODO`s that haven't been addressed yet 😅.

crates/pevm/Cargo.toml
```diff
@@ -5,23 +5,22 @@ edition = "2021"
```
hai-rise, line 5: This PR should not make any change to this file, let's revert 🙏.

crates/pevm/src/storage/rpc.rs
```diff
@@ -1,5 +1,11 @@
-use std::{fmt::Debug, future::IntoFuture, sync::Mutex, time::Duration};
+use std::{
+    fmt::Debug,
+    future::{Future, IntoFuture},
+    sync::{Mutex, OnceLock},
+    time::Duration,
+};
 
+use crate::{AccountBasic, EvmAccount, Storage};
```
hai-rise, line 8: Nit but can we keep this below and closer to `super::`?

crates/pevm/src/storage/rpc.rs
```diff
@@ -37,8 +44,7 @@ pub struct RpcStorage<N: Network> {
     cache_accounts: Mutex<ChainState>,
     cache_bytecodes: Mutex<Bytecodes>,
     cache_block_hashes: Mutex<BlockHashes>,
-    // TODO: Better async handling.
-    runtime: Runtime,
+    runtime: OnceLock<Runtime>,
```
hai-rise, line 47: Very nit but we can move this to the top now to let the caches flow to the bottom (now that there's no comment to separate them and the runtime).

crates/pevm/src/storage/rpc.rs
```diff
@@ -51,36 +57,47 @@ impl<N: Network> RpcStorage<N> {
...
-            runtime: Runtime::new().unwrap(),
+            runtime: OnceLock::new(),
+        }
+    }
+
+    #[inline(always)]
+    /// Awaits on a future using the current Tokio runtime if it exists,
+    /// or creates a new one (at most once) if it doesn't.
+    pub fn block_on<F: Future>(&self, future: F) -> F::Output {
+        if let Ok(runtime) = Handle::try_current() {
```
hai-rise, line 68: This should be `handle` not `runtime`.

crates/pevm/src/storage/rpc.rs
```diff
@@ -51,36 +57,47 @@ impl<N: Network> RpcStorage<N> {
...
+            task::block_in_place(|| runtime.block_on(future))
+        } else {
+            self.runtime
+                .get_or_init(|| Runtime::new().expect("Failed to create Tokio runtime"))
+                .handle()
```
hai-rise, line 73: We can call `block_on` directly on the `Runtime`.

crates/pevm/src/storage/rpc.rs
```diff
@@ -51,36 +57,47 @@ impl<N: Network> RpcStorage<N> {
             cache_accounts: Mutex::default(),
             cache_bytecodes: Mutex::default(),
             cache_block_hashes: Mutex::default(),
-            // TODO: Better error handling.
-            runtime: Runtime::new().unwrap(),
+            runtime: OnceLock::new(),
+        }
+    }
+
+    #[inline(always)]
```
hai-rise, line 64: Really like the implementation now! But to simplify we can remove this (we don't need to micro-optimize this module) and the comments as the code is pretty self-explanatory. 

If we were to add a comment I'd add that we currently only support a global multi-thread runtime.

hai-rise (changes requested):
I've left few more nits but we're very close now!

hai-rise (approved):
Awesome, we can now run `RpcStorage` in a (multithreaded) Tokio async context plus faster RPC test!

Issue #421, opened 2024-11-04
Hardcode Ethereum & OP chain ids instead of using `alloy-chains`


hai-rise:
Let's not declare an explicit dependency for such a simple use case.

PR #425, opened 2024-11-06
refactor: hardcode Ethereum & OP chain ids(#421)

byhashdong:
#421


hai-rise (approved):
Thank you! I'll get back to #409 soon™️, been a bit hectic lately 🥲.

Issue #420, opened 2024-11-04
CI: Add `cargo-machete`


hai-rise:
https://github.com/bnjbvr/cargo-machete

hai-rise:
Done in #426.

PR #426, opened 2024-11-06
CI: Add cargo-machete (#420)

byhashdong:
#420

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -15,6 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - uses: bnjbvr/cargo-machete@main
```
hai-rise, line 18: We should put this below `actions/cache` for it to get cached eventually? Otherwise, it'd add 45s to every run :sweat_smile:.

hai-rise (changes requested):

hai-rise (approved):
Nice, thanks!

PR #436, opened 2024-11-21
chore: add reoder_keys option format with taplo into ci

### LINKED ISSUE #433: Format TOML with `taplo` and `reorder_keys = true`
https://taplo.tamasfe.dev/configuration/formatter-options.html

We should also add `taplo` to CI.

mateusfigmelo:
Fixes #433

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
             ~/.cargo
             ./target
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-      
+          
       - uses: bnjbvr/cargo-machete@main
 
       - name: Lint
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -24,6 +24,12 @@ jobs:
             ./target
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
 
+      - name: Install Taplo
```
hai-rise, line 27: We should add this to `fmt.yml`!

hai-rise (changes requested):
I'm 50/50 about checking in a Taplo config file. @i1i1 what do you think?

.github/workflows/ci.yml
```diff
@@ -23,7 +23,7 @@ jobs:
             ~/.cargo
             ./target
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-
+
```
hai-rise, line 26: Please revert 🥲.

hai-rise (changes requested):

hai-rise (approved):
Nice!

Issue #351, opened 2024-09-14
Write more OP tests


hai-rise:
First, abstract Ethereum tests over any `Chain`:
- [ ] raw transfers
- [ ] erc20
- [ ] uniswap
- [ ] mixed
- [ ] beneficiary
- [ ] small blocks
- [ ] mainnet blocks

Then write more dedicated tests like with deposit transactions.

PR #437, opened 2024-11-27
build: bump deps

hai-rise:


i1i1 (commented):
Btw, WDUT of adding dependabot to repo. It can bump most of the dependencies (probably apart from alloy/reth and our git dependencies :thinking: 

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates

hai-rise:
> Btw, WDUT of adding dependabot to repo. It can bump most of the dependencies (probably apart from alloy/reth and our git dependencies 🤔
> 
> https://docs.github.com/en/code-security/dependabot/dependabot-version-updates

~~I hate dependabot with a passion.~~ Our crates are very sensitive to dependencies (both correctness and performance) so I'd prefer to bump most at once for a careful test & bench. Alloy is breaking API often so `cargo update` can easily fail to build. We also only need this every few weeks so no need to automate yet.

kien-rise (approved):
LGTM

Tested with a few of blocks on Optimism Mainnet.
Note that new blocks after Granite are failing.

```
Granite    approved    Wed Sep 11 16:00:01 UTC 2024 (1726070401) around block 125235812
```

hai-rise:
@i1i1 In all seriousness, things like `dependabot` will be helpful once we have #351 and #388 in place to automate the testing of each dependency bump. Until then, we can manually do it every few weeks, all dependencies together :).

PR #443, opened 2024-12-12
Speed up profiling and release profiles and add maxperf

i1i1:


mateusfigmelo (approved):
LGTM

hai-rise (changes requested):
Hmm, I don't think we need this many profiles for `pevm` as:
- The difference between `release` and `maxperf` should be much smaller than for the `rise` node and Reth?
- This is mainly a library so in practice build options would be driven by the binary that imports it anyway?

Maybe we can just add a `profiling` profile to replace `CARGO_PROFILE_BENCH_DEBUG` in the bench doc :thinking:.

Cargo.toml
```diff
@@ -3,10 +3,9 @@ default-members = ["crates/pevm"]
 members = ["bins/fetch", "crates/pevm"]
 resolver = "2"
 
-[profile.release]
```
hai-rise, line 6: We still need to keep these changes, mainly for the benchmarks to pick up max perf without the need to specify a profile.

Cargo.toml
```diff
@@ -3,10 +3,9 @@ default-members = ["crates/pevm"]
...
-codegen-units = 1
-lto = "fat"
-panic = "abort"
+[profile.profiling]
```
hai-rise, line 6: Can you replace `CARGO_PROFILE_BENCH_DEBUG` in the bench doc with this profile?

hai-rise (changes requested):

Cargo.toml
i1i1: I don't think it's worth it TBH for benchmarking. Compile time wise, it'll speed up compilation from 50s to 7s (at least on my machine).

My benchmarks with default release:
```
Independent Raw Transfers/Sequential time:   [97.991 ms 100.06 ms 102.26 ms]
Independent Raw Transfers/Parallel time:   [53.939 ms 55.938 ms 58.185 ms]
Independent ERC20/Sequential time:   [194.78 ms 197.20 ms 199.84 ms]
Independent ERC20/Parallel time:   [73.346 ms 75.232 ms 77.259 ms]
Independent Uniswap/Sequential time:   [752.41 ms 758.91 ms 765.59 ms]
Independent Uniswap/Parallel time:   [123.49 ms 125.89 ms 128.63 ms]
```
vs `lto=fat,codegen-units=1`:
```
Independent Raw Transfers/Sequential time:   [86.898 ms 88.229 ms 89.691 ms] (change: [-14.122% -11.823% -9.5278%])
Independent Raw Transfers/Parallel time:   [49.614 ms 50.574 ms 51.665 ms] (change: [-13.431% -9.5899% -5.6750%])
Independent ERC20/Sequential time:   [191.18 ms 194.58 ms 198.23 ms] (change: [-3.4492% -1.3305% +0.8373%])
Independent ERC20/Parallel time:   [71.877 ms 73.789 ms 75.965 ms] (change: [-5.2444% -1.9181% +1.9889%])
Independent Uniswap/Sequential time:   [724.53 ms 730.30 ms 736.37 ms] (change: [-4.9064% -3.7690% -2.5954%])
Independent Uniswap/Parallel time:   [122.51 ms 125.41 ms 128.48 ms] (change: [-3.4218% -0.3753% +2.8081%])
```

Sometimes I hit regressions even with LTO and codegen-units=1, but to me results are identical. So I don't think it would matter if we enable extra optimizations by default, as it will just slow down the compilation.

WDUT?

hai-rise: There are so many wrongs here...

1. We are benchmarking to find the max speed up and raw execution speed. Anything but max perf is wrong, nothing to do with convenience.
2. A `release` build under a minute is fast. If you need 7s for development use `dev` or `cargo check`.
3. -11.823%, -9.5899%, -1.3305%, -1.9181%, -3.7690%, -0.3753% are literally not identical. A 10% improvement could replace the whole scheduler in the Linux kernel. Of course it matters.

For context, the last time I played with these `codegen-units=1` made much of the difference. LTO is extremely sus (some up and down) but given that it is on for Reth and `rise`'s `maxperf` builds, it makes sense for us to benchmark and optimize for it.

i1i1: > 1. We are benchmarking to find the max speed up and raw execution speed. Anything but max perf is wrong, nothing to do with convenience.

True, but I don't think it changes the picture if we don't hit *some* compiler optimizations. Those are nits, that is why they are not enabled by default.

> 2. A release build under a minute is fast. If you need 7s for development use dev or cargo check.

True, but my concern here is that `lto=fat,codegen-units=1` compile times grow exponentially with the number of dependencies.

> 3. -11.823%, -9.5899%, -1.3305%, -1.9181%, -3.7690%, -0.3753% are literally not identical. A 10% improvement could replace the whole scheduler in the Linux kernel. Of course it matters.

Yeah, but those are flaky and not across all workloads. (I don't have numbers by hand here, but I got similar results to default release on my machine)

hai-rise: > True, but I don't think it changes the picture if we don't hit some compiler optimizations.

What we think doesn't matter. Let's not pretend we understand how these Rust and LLVM compiler optimizations work. That is why we have benchmarks to have numbers easy enough to evaluate.

> Those are nits, that is why they are not enabled by default.

Typical strawman argument. Not all Rust projects are set out to become the fastest at something, most aren't even that performance-critical. Yeah, let's use them as the standard instead of something like Actix:
https://github.com/actix/actix-web/blob/8115c818c15792a6edab01f939f049b6c0e38792/Cargo.toml#L28-L31

> True, but my concern here is that lto=fat,codegen-units=1 compile times grow exponentially with the number of dependencies.

We should be concerned about throughput and latency, not sacrificing benchmark quality to bring compile time from 50s to 7s when they run for 45-60 minutes. We are trying to replace the biggest dependency (`revm`) with an inline implementation. Genuinely not sure where this imaginary concern came from.

> Yeah, but those are flaky and not across all workloads. (I don't have numbers by hand here, but I got similar results to default release on my machine)

You pasted changes up to 11% so I wouldn't trust your definition of "similar" until you do more pevm optimizations and micro-benchmarks in general. 3% is the flaky range for a complicated E2E setup like for RISE devnet. For this much more stable environment, it is clear regression. Criterion would show red for even ~1.5% regression.

Cargo.toml
```diff
@@ -8,6 +8,10 @@ codegen-units = 1
 lto = "fat"
 panic = "abort"
 
+[profile.profiling]
```
hai-rise, line 11: While we're at it let's `strip = true` in `release` and `strip = false` here to match `rise` and Reth.

hai-rise (changes requested):
One last nit.

hai-rise (approved):
Nice!

PR #444, opened 2024-12-13
Simplify receipts root calculation

i1i1:


crates/pevm/src/chain/optimism.rs
```diff
@@ -189,45 +187,34 @@ impl PevmChain for PevmOptimism {
...
+                            deposit_nonce: (spec_id >= SpecId::CANYON).then_some(account.nonce - 1),
+                            deposit_receipt_version: (spec_id >= SpecId::CANYON).then_some(1),
+                        };
+                        OpReceiptEnvelope::Deposit(receipt.with_bloom())
+                    }
+                }
             })
-            .collect();
+            .enumerate()
+            .map(|(index, receipt)| (alloy_rlp::encode_fixed_size(&index), receipt.encoded_2718()))
```
kien-rise, line 215: Just curious, how does `.encoded_2718` accept zero arguments? 

I thought it always accept one argument? `receipt.encode_2718(&mut value_buffer);`

kien-rise (approved):
No blocking issues. But remember to run some tests, especially for optimism side.

crates/pevm/src/chain/optimism.rs
i1i1: Yes! You right, but this method is `.encoded_2718` :)

It is basically:
```rust
fn encoded_2718(&self) -> Vec<u8> {
    let mut buf = Vec::with_capacity(self.encode_2718_len());
    self.encode_2718(&mut buf);
    buf
}
```

crates/pevm/src/chain/ethereum.rs
```diff
@@ -196,19 +193,11 @@ impl PevmChain for PevmEthereum {
                     TxType::Eip4844 => ReceiptEnvelope::Eip4844(receipt),
                     TxType::Eip7702 => ReceiptEnvelope::Eip7702(receipt),
                 }
```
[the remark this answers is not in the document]
i1i1: It doesn't actually implement any kind of `From` which is sad:
https://reth.rs/docs/reth/rpc/types/enum.ReceiptEnvelope.html#trait-implementations

crates/pevm/src/chain/optimism.rs
kien-rise: oh, cool. I didn't notice the difference between `encoded_2718` and  `encode_2718` (the letter `d`)

i1i1: Also, it should be faster (but it's not worth it here), as we will allocate precisely once (`Vec::with_capacity`) and it won't reallocate

hai-rise (approved):
Nice!

PR #456, opened 2025-01-10
chore: bump deps

hai-rise:


crates/pevm/src/chain/ethereum.rs
```diff
@@ -58,7 +51,7 @@ fn get_ethereum_gas_price(tx: &TxEnvelope) -> Result<U256, EthereumTransactionPa
 impl PevmChain for PevmEthereum {
     type Transaction = alloy_rpc_types_eth::Transaction;
     type Envelope = TxEnvelope;
-    type BlockSpecError = EthereumBlockSpecError;
+    type BlockSpecError = ();
```
i1i1, line 54: I'm a bit late to the party, but this could be [`std::convert::Infallible`](https://doc.rust-lang.org/nightly/std/convert/enum.Infallible.html) which makes `Result::Err` not constructable and makes functions always return `Result::Ok`

hai-rise: Ah yes, thanks. More care and justice for `pevm` soon™️!

PR #459, opened 2025-02-11
Allow call new contract code in one block

jackwener:


hai-rise:
Closing as we're going through a full redesign :pray:.

PR #501, opened 2026-04-26
chore: switch ethereum state tests from ethereum/tests to EEST

hai-rise:
Replaces the deprecated ethereum/tests git submodule with fixtures downloaded from ethereum/execution-spec-tests (the current standard, now maintained in ethereum/execution-specs). Run once before testing:

  bash scripts/fetch-eest-fixtures.sh

Key changes:
- Remove the ethereum/tests submodule
- Add scripts/fetch-eest-fixtures.sh to download fixtures_stable.tar.gz
- Rewrite main.rs to walk state_tests/ instead of GeneralStateTests/, map all EEST exception strings (with pipe-separated OR support), and skip the same create-collision tests revm can't handle correctly
- Add fixtures directory to .gitignore

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,5 +36,5 @@ jobs:
         # Running our heavy tests in parallel would congest resources.
         # Each test still executes parallelly anyway.
         run: |
-          git submodule update --init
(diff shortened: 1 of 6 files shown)


tu-doan:
benchmark

Issue #502, opened 2026-05-07
Implement a Parallel-EVM-aware Journal


hai-rise:
Then put it in here:
https://github.com/risechain/pevm/blob/0734862b1997a7756ecd50c980342ff9bf186bf3/crates/pevm/src/chain.rs#L60

Towards #382.

duyquang6:
I'll work on this

PR #503, opened 2026-05-12
feat: journal

duyquang6:
Downstream journal, derive write inside Journal 
- storage writeset (in testing)
- account writeset (TBU)
- code writeset (TBU)
- selfdestructed

Status: this PR regressed around 2% on some mainnet blocks, still working on it
reference: #502


crates/pevm/src/chain.rs
```diff
@@ -57,7 +59,7 @@ pub trait PevmChain: Debug {
             Context: ContextTr<
                 Db = DB,
                 Tx = Self::EvmTx,
-                Journal: JournalTr<State = EvmState>,
+                Journal: JournalTr<State = EvmState> + PevmJournalExt,
```
hai-rise: Ah, just rename `Journal` to `PevmJournal` and put `Journal = PevmJournal<DB>` concretely here.

Like with Reth in RISE nodes, we should rewrite as many traits as concrete types downstream as possible.

hai-rise (changes requested):
One quick specific comment, and 2 general ones:
- Let's push directly to the main repository for ease of pulling for other team members.
- Let's be careful with performance as Parallel EVM is very sensitive even to small low-level changes.

duyquang6:
benchmark

PR #510, opened 2026-05-26
fix: require full block_size validations before scheduler completion

datnguyencse:
With NeedValidation = 1 (distinct bit values), lazy and first transactions
auto-validate in finish_execution, incrementing num_validated without going
through the validation queue. This left min_validation_idx > 0, making the
completion threshold block_size - min_validation_idx < block_size.

Threads could exit as soon as that lower threshold was met, while the
remaining high-conflict non-lazy tail transactions were still mid
re-execution. Those transactions then ran with a shrinking thread pool,
degrading parallelism on contentious workloads.

Fix: require num_validated >= block_size unconditionally. Auto-validations
still count toward the total, so pure lazy blocks are unaffected; all
threads now stay alive until every transaction's conflict resolution is
complete.

Benchmark:
- NeedValidation = 0
```
Mainnet:
Average: x1.97
Max: x3.97
Min: x0.66

Gigagas:
Raw: 47.715 ms
ERC20: 54.146 ms
Uniswap: 25.270 ms
```
- NeedValidation = 1
```
Average: x1.93
Max: x3.92
Min: x0.67

Raw: 47.287 ms
ERC20: 54.055 ms
Uniswap: 25.179 ms
```
- This PR:
```
Average: x1.97
Max: x3.94
Min: x0.66

Raw: 47.526 ms
ERC20: 53.470 ms
Uniswap: 25.217 ms
```

### DIFF
--- a/crates/pevm/src/pevm.rs
+++ b/crates/pevm/src/pevm.rs
@@ -181,7 +181,7 @@ impl Pevm {
         }
 
         let block_size = txs.len();
-        let scheduler = Scheduler::new(block_size);
+        let scheduler = Scheduler::new(block_size, concurrency_level.into());
 
         let mv_memory = chain.build_mv_memory(&block_env, &txs);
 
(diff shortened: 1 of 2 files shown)


tu-doan:
benchmark

tu-doan:
benchmark

PR #514, opened 2026-06-04
feat: collision-resistant memory-location hashing via seed + u64 tag

datnguyencse:
## Problem

`MemoryLocationHash = u64` is used **as the `MvMemory` map key itself**, so it doubles as both the bucket key and the location's identity. Two distinct locations that collide in the `u64` share one entry:

- **Targeted:** FxHash is non-cryptographic and algebraically invertible — an attacker can craft a storage slot that collides with another location for ~free, making a read return the wrong value (which then validates cleanly) → **wrong state / chain fork**, or flood one bucket → **DoS**.
- **Random:** any 64-bit key collides at the ~2³² birthday bound (~10⁻¹⁰–10⁻¹² per block).

## Fix — two cheap, independent levers

1. **Per-execution random seed** mixed into every location hash & tag. Unknown to transaction submitters until the block runs, so collisions cannot be aimed at this block. Derived from `RandomState` (no new dependency) and stored on `MvMemory`.
2. **Per-entry verification tag** — a second, independently seeded fingerprint (`!seed`) stored in each `MemoryEntry`. Reads skip entries whose tag differs from the location being read, so a collision is only harmful if **both** the hash and the tag collide (~2⁻¹²⁸). This restores the key-comparison a normal `HashMap` does, without widening the `u64` map key.

The seed never escapes into results (state is keyed by real `Address`es), so **output is identical for any seed**.

## Cost

Read path (common, no-collision case): one extra `u64` compare + one extra hash per location. Per block: one `RandomState::new()`.

## Residual (documented, adversarially unreachable)

`ReadSet` / validation stay keyed by the bare `u64` hash, so two colliding locations touched by one tx can still conflate there — but that only ever forces a **spurious re-execution (liveness), never wrong state**, and the secret seed makes it impossible to trigger on purpose.

## Verification

- `cargo clippy --workspace --all-targets` — clean
- `cargo test --workspace --release -- --test-threads=1` — **all green**, including the full `ethereum/tests` state-test suite, RISE/mainnet/uniswap blocks (these compare pevm output to expected/sequential, proving the change is output-preserving)

Benchmarks (`gigagas`, `mainnet`) pending — opening as draft for perf review.

### DIFF
--- a/crates/pevm/src/chain.rs
+++ b/crates/pevm/src/chain.rs
@@ -25,7 +25,7 @@ use revm::{
 };
 use smallvec::SmallVec;
 
-use crate::{MemoryLocationHash, PevmTxExecutionResult, mv_memory::MvMemory};
+use crate::{LocationTag, MemoryLocationHash, PevmTxExecutionResult, mv_memory::MvMemory};
 
 /// The error type of [`PevmChain::calculate_receipt_root`]
 #[derive(Debug, Clone)]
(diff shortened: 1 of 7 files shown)


datnguyencse:
benchmark

PR #515, opened 2026-06-04
feat: collision-resistant memory-location hashing via seed + u128 key

datnguyencse:
## Problem

`MemoryLocationHash = u64` is used **as the `MvMemory` map key itself**, so it doubles as both the bucket key and the location's identity. Two distinct locations that collide in the `u64` share one entry:

- **Targeted:** FxHash is non-cryptographic and algebraically invertible — an attacker can craft a storage slot that collides with another location for ~free, making a read return the wrong value (which then validates cleanly) → **wrong state / chain fork**, or flood one bucket → **DoS**.
- **Random:** any 64-bit key collides at the ~2³² birthday bound (~10⁻¹⁰–10⁻¹² per block).

## Fix — two independent levers

1. **Widen `MemoryLocationHash` to `u128`** (built from two decorrelated FxHash passes, `seed` and `!seed`, forming the low/high 64 bits). The map's own 128-bit key comparison becomes a reliable identity check, so random collisions drop to the ~2⁻⁶⁴ birthday bound (effectively never). `IdentityHasher` projects the key to its low 64 bits for bucketing, so DashMap sharding cost is unchanged.
2. **Per-execution random seed** mixed into the hash. Unknown to transaction submitters until the block runs, so collisions cannot be aimed at this block. Derived from `RandomState` (no new dependency), stored on `MvMemory`.

The seed never escapes into results (state is keyed by real `Address`es), so **output is identical for any seed**.

## Cost

Hashing does two FxHash passes per location instead of one (~2× a cheap op, negligible vs EVM execution); keys widen `u64`→`u128` (+8 bytes per live key); bucketing/sharding unchanged. One `RandomState::new()` per block.

## Verification

- `cargo clippy --workspace --all-targets` — clean
- `cargo test --workspace --release -- --test-threads=1` — **all green**, including the full `ethereum/tests` state-test suite, RISE/mainnet/uniswap blocks (these compare pevm output to expected/sequential, proving the change is output-preserving)

Benchmarks (`gigagas`, `mainnet`) pending — opening as draft for perf review. An alternative implementation (seed + per-entry tag, keeping the `u64` key) exists for comparison.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/pevm/src/chain.rs
+++ b/crates/pevm/src/chain.rs
@@ -139,18 +139,21 @@ pub trait PevmChain: Debug {
         true
     }
 
-    /// Build [`MvMemory`]
-    fn build_mv_memory(&self, _block_env: &BlockEnv, txs: &[Self::EvmTx]) -> MvMemory {
-        MvMemory::new(txs.len(), [], [])
+    /// Build [`MvMemory`]. `seed` is the per-execution random seed that must be
+    /// mixed into every location hash & tag (see [`crate::hash_deterministic`]).
+    fn build_mv_memory(&self, _block_env: &BlockEnv, txs: &[Self::EvmTx], seed: u64) -> MvMemory {
+        MvMemory::new(seed, txs.len(), [], [])
     }
 
-    /// Get rewards (balance increments) to beneficiary accounts, etc.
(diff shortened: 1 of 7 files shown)


datnguyencse:
benchmark

PR #520, opened 2026-08-18
fix(vm): track EIP-7702 code changes

### LINKED ISSUE #518: EIP-7702 re-delegation can leave a stale code hash in MV memory
## Summary

When an EIP-7702 authority is delegated more than once in the same block, the parallel executor may keep using the first delegation after a later transaction changes it.

The write-set construction currently treats code as new only when the account previously had no code:

```rust
let has_code = !account.info.is_empty_code_hash();
let is_new_code = has_code
    && read_account.is_none_or(|(_, code_hash)| code_hash.is_none());
```

This detects an empty-to-non-empty transition, but not a non-empty-to-different-non-empty transition.

## Reproduction scenario

Under a spec that enables EIP-7702:

1. Transaction 10 delegates authority A to contract X.
2. Transaction 20 re-delegates authority A to contract Y.
3. Transaction 30 calls authority A.

Transaction 30 should execute the code from Y. During parallel execution, transaction 20 reads a non-empty code hash for X, so `is_new_code` is false. As a result, the updated `CodeHash` entry and bytecode for Y are not published to multi-version memory. A later read can therefore resolve the stale code hash for X.

This can make parallel execution diverge from sequential execution.

## Expected behavior

A code entry should be published whenever the post-state code hash differs from the code hash read by the transaction, not only when code appears for the first time.

Conceptually, the condition should compare the previous and current hashes:

```rust
let code_changed = has_code
    && account.info.code.is_some()
    && read_account.is_none_or(|(_, code_hash)| {
        *code_hash != Some(account.info.code_hash)
    });
```

The updated code hash and bytecode should then be included in the write set when `code_changed` is true.

## Suggested regression coverage

A block-level parallel-vs-sequential test would be useful for:

- EOA to delegation X
- delegation X to delegation Y in the same block
- delegation X to cleared delegation
- delegation X to cleared delegation to delegation Y

The re-delegation test should include a later call whose state change makes it observable whether X or Y was executed.

pekkahcut:
Fixes #518

Track EIP-7702 code changes when an account is re-delegated to a different target in the same block. Previously, the write set only treated code as new when there was no previous code hash, which could leave a stale code hash in MV memory.

This compares the current code hash with the previously read code hash and writes the updated contract when they differ.

Tests:
- `cargo test -p pevm --lib`
- `cargo fmt --all`
- `git diff --check`

### DIFF
--- a/crates/pevm/src/vm.rs
+++ b/crates/pevm/src/vm.rs
@@ -596,11 +596,14 @@ impl<'a, S: Storage, C: PevmChain> Vm<'a, S, C> {
                         let read_account = ctx.db().read_accounts.get(&account_location_hash);
 
                         let has_code = !account.info.is_empty_code_hash();
-                        let is_new_code = has_code
-                            && read_account.is_none_or(|(_, code_hash)| code_hash.is_none());
+                        let code_changed = has_code
+                            && account.info.code.is_some()
+                            && read_account.is_none_or(|(_, code_hash)| {
+                                *code_hash != Some(account.info.code_hash)
+                            });
 
                         // Write new account changes
-                        if is_new_code
+                        if code_changed
                             || read_account.is_none()
                             || read_account.is_some_and(|(basic, _)| {
                                 basic.nonce != account.info.nonce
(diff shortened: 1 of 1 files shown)


pekkahcut:
Closing in favor of #519, which addresses the same issue and includes broader regression coverage. Thanks!

pekkahcut:
Closing in favor of #519, which addresses the same issue and includes broader regression coverage. Thanks!

